PR #11, opened 2022-06-27
Support for ENUM type

suranjan:
With this change the ENUM type is sent as STRING and values are also sent as STRING. There is a test, however that doesn't work because the docker image doesn't have yugabytedb version that supports ENUM type yet


src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
```diff
@@ -220,6 +228,31 @@ public void testVerifyValue() throws Exception {
...
+
+        // 3 because there are 3 enum values in the enum type
+        final long recordsCount = 3;
+
+        awaitUntilConnectorIsReady();
+
+        // 3 records will be inserted in the table test_enum
+        insertEnumRecords();
+
+        CompletableFuture.runAsync(() -> verifyPrimaryKeyOnly(recordsCount))
```
vrajat, line 250: Is there an assert to ensure enum value is correct in the source record?

vrajat (commented):
Does the test check for enum values?

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
vaibhav-yb: We can add one anytime.

vaibhav-yb (approved):
Looks good to me. The test can be changed to use a DB version with ENUM support in the next iteration.

suranjan:
> Does the test check for enum values?

yes, added the test.

PR #18, opened 2022-07-04
Upgrade Connector to Debezium Core version 1.9.4.Final

vrajat:


src/main/java/io/debezium/connector/yugabytedb/YBPartition.java
```diff
@@ -11,6 +11,10 @@ public class YBPartition implements Partition {
 
     private final String tabletId;
 
+    public YBPartition() {
+        this.tabletId = "";
```
vaibhav-yb, line 15: With this, there is a possibility that in a certain code flow if we have a condition where we check for tabletId to be null then having this will bypass that condition. We need to be sure that we are not violating that. eg.
```java
if (partition.getTabletId() == null ) {
  // do something
} else {
  // do something else - with the above change, we will end up coming to this block all the time
}
```

src/main/java/io/debezium/connector/yugabytedb/connection/pgproto/YbProtoMessageDecoder.java
```diff
@@ -24,6 +24,7 @@
 import io.debezium.connector.yugabytedb.proto.PgProto.Op;
 import io.debezium.connector.yugabytedb.proto.PgProto.RowMessage;
 import io.debezium.util.Collect;
+import org.yb.cdc.CdcService;
```
vaibhav-yb, line 27: Seems like this is the only change in this file. If so then it can be removed if it's not being used.

vaibhav-yb (changes requested):
I will need another pass at some of the critical files like `YugabyteDBStreamingChangeEventSource` and YugabyteDBEventDispatcher` etc, but the rest of the changes look good.

Btw, the latest stable in DBZ is `1.9.5.Final` in case we want to use that.

vaibhav-yb (approved):

src/main/java/io/debezium/connector/yugabytedb/YBPartition.java
vrajat: This is a good point. I checked PostgresConnector and checked that it shouldn't have the default constructor. I removed and then have fixed a bunch of APIs to calculate previous offsets. Thanks for pointing this out.

PR #20, opened 2022-07-11
log LSN when we receive 0 records in transactional block

adithya-yb:
Log the LSN of the transaction which had 0 records in GetChanges


vrajat (approved):

adithya-yb:
Log transactionId and LSN when we receive records for a transactional block in GetChanges

PR #22, opened 2022-07-19
Add docker plugin to push image to quay using maven lifecycles

vaibhav-yb:
This PR aims to integrate the maven docker plugin so that builds and the respective images can automatically be pushed to quay to be further used by users.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,8 @@
 # On your terminal, run the following to build the image:
 #  - mvn clean package -Dquick
 
-# The base image contains the following drivers and connectors prepopulated:
+# The base image is derived from debezium/connect:1.7 and contains the following drivers 
+# and connectors prepopulated:
 #  - JDBC Sink Connector v10.2.5
 #  - YugabyteDB JDBC Driver v42.3.5-yb-1
 #  - MySql JDBC Driver v8.0.21
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,11 @@
+# On your terminal, run the following to build the image:
+#  - mvn clean package -Dquick
+
 # The base image contains the following drivers and connectors prepopulated:
 #  - JDBC Sink Connector v10.2.5
 #  - YugabyteDB JDBC Driver v42.3.5-yb-1
 #  - MySql JDBC Driver v8.0.21
-FROM quay.io/yugabyte/connect-base-yb:0.1
+FROM quay.io/yugabyte/connect-base-yb:0.2
 
 # Create the directories for the connectors to be placed into
 ENV KAFKA_CONNECT_YB_DIR=$KAFKA_CONNECT_PLUGINS_DIR/debezium-connector-yugabytedb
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -467,6 +411,42 @@
...
+                        </image>
+                    </images>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>build</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>build</goal>
+                            <goal>push</goal>
```
vrajat, line 438: If you remove this, then it won't push during `package`.

pom.xml
```diff
@@ -505,7 +485,7 @@
             <properties>
                 <skipITs>true</skipITs>
                 <maven.test.skip>true</maven.test.skip>
-                <docker.skip>true</docker.skip>
+                <!-- <docker.skip>false</docker.skip> -->
```
vrajat, line 488: For `quick` remove this line. We want `mvn package -Dquick` to create a docker image.

pom.xml
```diff
@@ -521,7 +501,7 @@
                 </property>
             </activation>
             <properties>
-                <docker.skip>true</docker.skip>
+                <!-- <docker.skip>true</docker.skip> -->
```
vrajat, line 504: I don't even know when this is used. If you don't use it regularly, I suggest removing the whole profile.

pom.xml
```diff
@@ -535,7 +515,7 @@
             <properties>
                 <skipITs>true</skipITs>
                 <maven.test.skip>true</maven.test.skip>
-                <docker.skip>true</docker.skip>
+                <!-- <docker.skip>true</docker.skip> -->
```
vrajat, line 518: We do want docker images to be created during releases. So this line should be removed.

vrajat (changes requested):

Dockerfile
```diff
@@ -0,0 +1,15 @@
+# The base image contains the following drivers and connectors prepopulated:
+#  - JDBC Sink Connector v10.2.5
+#  - YugabyteDB JDBC Driver v42.3.5-yb-1
+#  - MySql JDBC Driver v8.0.21
+FROM quay.io/yugabyte/connect-base-yb:0.1
```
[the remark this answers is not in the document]
vrajat: Also we need the ability to build from different base images. `1.6` is too old. Current one should be `1.7` and later move to `1.9`.

pom.xml
```diff
@@ -19,7 +19,7 @@
     <properties>
         <!-- Debezium parent -->
         <version.debezium>1.7.0.Final</version.debezium>
-        <version.debezium.connector.yb>1.3.8-BETA</version.debezium.connector.yb>
+        <version.debezium.connector.yb>1.7.0.8</version.debezium.connector.yb>
```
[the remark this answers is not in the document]
vrajat: Change `${version.debezium.connector.yb}` to `${project.version}` in line 426

vrajat (approved):

PR #26, opened 2022-08-11
Snapshot wip

suranjan:
Support for snapshot.mode=initial.
We don't use SnapshotSource as the APIs are same. 
We use internal APIs for snapshot that is provide new OpId(-1,-,1, [], -1,-1,-1) as the from_op_id and the server will take a snapshot and stream the changes.
After snapshot, the connector will stop.

The connector needs to be started with Snapshot.mode=never for further streaming changes.


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnector.java
```diff
@@ -211,11 +211,13 @@ private YBClient getYBClient(String hostAddress, long adminTimeout, long opTimeo
     @Override
     public void stop() {
         this.props = null;
-        try {
+        if (this.ybClient != null) {
+          try {
             ybClient.close();
-        }
-        catch (Exception e) {
+          }
```
vrajat, line 217: These look like white space changes? Can it be reverted?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -349,11 +372,12 @@ private void getChanges2(ChangeEventSourceContext context,
         // Note that the entry will be created only when a BEGIN block is encountered.
         Map<String, Integer> recordsInTransactionalBlock = new HashMap<>();
 
-        LOGGER.debug("The init tabletSourceInfo is " + offsetContext.getTabletSourceInfo());
+        LOGGER.info("The init tabletSourceInfo is " + offsetContext.getTabletSourceInfo());
```
vrajat, line 375: The same information is logged twice. First time in line 359 above.

vrajat (changes requested):

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
```diff
@@ -269,6 +277,26 @@ public void testRecordConsumption() throws Exception {
...
+  @Test
+  public void testSnapshotRecordConsumption() throws Exception {
+    TestHelper.dropAllSchemas();
+    TestHelper.executeDDL("yugabyte_create_tables.ddl");
+    final int recordsCount = 5000;
+    // insert rows in the table t1 with values <some-pk, 'Vaibhav', 'Kushwaha', 30>
+    insertBulkRecords(recordsCount);
+
+    String dbStreamId = TestHelper.getNewDbStreamId("yugabyte", "t1");
+    Configuration.Builder configBuilder = TestHelper.getConfigBuilder("public.t1", dbStreamId);
```
vaibhav-yb, line 289: I do not see changes to this function and by default this returns a Configuration with snapshot mode set as never. Can you please check if this test works?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -116,19 +115,30 @@ public YugabyteDBStreamingChangeEventSource(YugabyteDBConnectorConfig connectorC
...
         // replication slot could exist at the time of starting Debezium so
         // we will stream from the position in the slot
         // instead of the last position in the database
-        boolean hasStartLsnStoredInContext = offsetContext != null;
+        // Get all partitions
+        // Get
+        Set<YBPartition> partitions = new YugabyteDBPartition.Provider(connectorConfig).getPartitions();
+        boolean hasStartLsnStoredInContext = offsetContext != null && !offsetContext.getTabletSourceInfo().isEmpty();
 
+        // LOGGER.info("SKSK The offset context is " + offsetContext + " partition is " + partition);
```
vaibhav-yb, line 126: This needs to be removed.

I guess we should also remove other stale comments like this which we may not need later to avoid confusion.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -334,6 +344,19 @@ private void getChanges2(ChangeEventSourceContext context,
...
+        // final String tabletId = entry.getValue();
+        // offsetContext.initSourceInfo(tabletId, this.connectorConfig);
+        //
+        // offsetContext.getSourceInfo(tabletId).updateLastCommit(offsetContext.lastCompletelyProcessedLsn());
+        //
+        // if (offsetContext.lsn(tabletId).equals(new OpId(0, 0, null, 0, 0))) {
+        // offsetContext.getSourceInfo(tabletId)
+        // .updateLastCommit(new OpId(-1, -1, "".getBytes(), -1, 0));
+        // }
+        // }
```
vaibhav-yb, line 358: Extra comments.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
vaibhav-yb: Also, do we need this information at the info level log? If not, then we should keep it at the debug level only.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -116,19 +115,30 @@ public YugabyteDBStreamingChangeEventSource(YugabyteDBConnectorConfig connectorC
...
+            }
+            else {
+                offsetContext = YugabyteDBOffsetContext.initialContext(connectorConfig, connection, clock, partitions);
+            }
         }
+        /*
+         * if (snapshotter.shouldSnapshot()) {
+         * getSnapshotChanges();
+         * }
+         */
```
vaibhav-yb, line 141: Extra comments.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -116,19 +115,30 @@ public YugabyteDBStreamingChangeEventSource(YugabyteDBConnectorConfig connectorC
 
     @Override
     public void execute(ChangeEventSourceContext context, YugabyteDBPartition partition, YugabyteDBOffsetContext offsetContext) {
-        if (!snapshotter.shouldStream()) {
-            LOGGER.info("Streaming is not enabled in correct configuration");
-            return;
-        }
-
         // replication slot could exist at the time of starting Debezium so
```
vaibhav-yb, line 118: This comment will also be confusing because it talks about how the Debezium PG connector operates with replication slots which is not true for us, so maybe we can completely remove it or add YB related info here.

vaibhav-yb (changes requested):
I guess, some cleanup is required at the top level. The rest of the changes look good to me. 

Apart from this, the code formatting is inconsistent (at some places its YB standard, at others its the one from DBZ standard), but we need not worry about this since we are planning to change the auto formatter properties to match the YB standards soon so that will be taken care of.

vrajat:
Tests seem to be broken. One of the tests goes into an infinite loop. I couldn't figure out which one. Can the test be fixed or disabled?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
vaibhav-yb: Update: This log statement meant to print `offsetContext.getTabletSourceInfo()` - one before updating the `offsetContext` and another after updating it. Updating the PR with the necessary changes.

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
vaibhav-yb: Update: Fixed the issue in the test.

PR #27, opened 2022-08-19
Restrict one read from YugabyteDB per iteration

vrajat:
ADD `cdc.poll.limit` configuration option. Default is FALSE.
If the option is set to true, then the connector will not call
getChanges until the queue has been drained by the consumer.


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -386,6 +397,8 @@ private void getChanges2(ChangeEventSourceContext context,
                                 table, streamId, tabletId,
                                 cp.getTerm(), cp.getIndex(), cp.getKey(), cp.getWrite_id(), cp.getTime(), schemaStreamed.get(tabletId));
 
+                        LOGGER.info("Processing {} records from getChanges call",
```
vaibhav-yb, line 400: This log should also be at the `debug` level otherwise they will clog up the whole log file.

vaibhav-yb (changes requested):

vaibhav-yb (approved):

PR #38, opened 2022-09-05
Implementing the transition from snapshot to streaming

vaibhav-yb:
With this change, the user would not need to change the snapshot mode from `initial` to `never` in order to further stream the changes.

Now, if the connector is deployed with `snapshot.mode=initial`, it would first take the snapshot and then would automatically switch to streaming new changes in the source database.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
@@ -326,8 +326,8 @@ public class YugabyteDBStreamingChangeEventSource implements
                         LOGGER.debug("Snapshot completed for all the tablets! Stopping.");
                         if (!snapshotter.shouldStream()) {
                             // This block will be executed in case of initial_only mode
-                            LOGGER.info("Snapshot completed for initial_only mode, you can stop the connector now");
-                            break;
+                            LOGGER.info("Snapshot completed for initial_only mode, stopping the connector now");
+                            return;
                         }
                       }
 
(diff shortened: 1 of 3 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -323,16 +323,31 @@ private void getChanges2(ChangeEventSourceContext context,
...
+                        LOGGER.debug("Snapshot completed for all the tablets! Stopping.");
+                        if (!snapshotter.shouldStream()) {
+                            // This block will be executed in case of initial_only mode
+                            LOGGER.info("Snapshot completed for initial_only mode, stopping the connector now");
+                            return;
+                        }
                       }
+
+
+                      OpId cp = snapshotter.shouldSnapshot() ? offsetContext.snapshotLSN(tabletId) : offsetContext.lsn(tabletId);
```
vrajat, line 335: Is the ternary operator required here as `cp` is set again in line 340 if snapshot is completed. Will it suffice to only set it to `snapshotLSN` in this line?

vaibhav-yb: Actually, yes. If we remove the ternary operator then in case of `snapshot.mode=never`, we will never reach to line 340 and we will end up taking the snapshot of the table.

Note that I would simply call this a "hack" to accommodate the snapshot and streaming in a single class, future plan is to have the logic in separate classes to be in sync with Debezium's way of things.

vrajat:
Will there be a corresponding test in `cdcsdk-testing` ?

vaibhav-yb:
@vrajat Yes, once we formally merge this PR, I will also add a test in cdcsdk. However, I started facing some blockers when I was making some changes to this PR so we'll need to wait till midnight today.

vrajat (approved):

PR #42, opened 2022-09-09
Refactoring classes for snapshot and streaming

vaibhav-yb:
This PR contains the diff for the following:
* Refactoring `YugabyteDBStreamingChangeEventSource` class to move the logic for snapshot into `YugabyteDBSnapshotChangeEventSource`
* Addition of a new configuration parameter `snapshot.again` - this parameter will indicate the connector to take the snapshot again after a restart on a tablet on which it is taken already and the default value for this parameter is `false`. In case the user sets this to `true`, the connector will go for a snapshot for the specified table after restarts.
* Addition of tests to verify the changes made in this diff.


vrajat (commented):
A lot of the code commented. Can they deleted instead?

vaibhav-yb:
> A lot of the code commented. Can they deleted instead?

@vrajat Yes, I will be removing them before approaching for final review. Just wanted to keep them in case things go sideways 😄

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -1,5 +1,189 @@
...
+
+  /**
+   * Helper function to get the Debezium style TableId of a table from table UUID
+   * @param ybClient the {@link YBClient} instance
+   * @param table the {@link YBTable} instance
+   * @return the {@link TableId}
+   * @throws Exception if a {@link YBTable} cannot be opened by the client
+   */
+  public static TableId getTableIdFromYbTable(YBClient ybClient, YBTable table) throws Exception {
+    ListTablesResponse resp = ybClient.getTablesList(table.getName());
```
vaibhav-yb, line 180: To use the `TableInfo` object, we need to make a RPC call to the server to give us all the tables (with the specified table name ofcourse). We are using this way because we need the `pgSchemaName` and `TableInfo` is the only class which has access to `pgSchemaName`.

Now I think this could cause too many calls to the server from the connector.

The other approach I can think of is to add a pgSchemaName property to the `YBTable` object - if we follow this way, then since we already have the cached `YBTable` objects for each tableId we will ultimately reduce our dependency on the server.

However, we can still cache this `ListTablesResponse` once with the information of all the tables in the cluster and then can access it as and when needed. But with this, I don't think storing unnecessary information (which we are not going to use often) on the connector seems to be a good idea.

@vrajat @suranjan What are your thoughts on this?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -34,33 +50,61 @@
...
 
     public YugabyteDBSnapshotChangeEventSource(YugabyteDBConnectorConfig connectorConfig,
                                                YugabyteDBTaskContext taskContext,
-                                               Snapshotter snapshotter,
+                                               Snapshotter snapshotter, YugabyteDBConnection connection,
                                                YugabyteDBSchema schema, YugabyteDBEventDispatcher<TableId> dispatcher, Clock clock,
                                                SnapshotProgressListener snapshotProgressListener) {
         super(connectorConfig, snapshotProgressListener);
-        this.connectorConfig = connectorConfig;
+        this.ybConnectorConfig = connectorConfig;
```
vrajat, line 77: Why is the name change required?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -34,33 +50,61 @@
...
+            Duration.ofMillis(connectorConfig.connectorRetryDelayMs()), Clock.SYSTEM);
+
+        this.yugabyteDbTypeRegistry = taskContext.schema().getTypeRegistry();
     }
 
     @Override
     public SnapshotResult<YugabyteDBOffsetContext> execute(ChangeEventSourceContext context, YBPartition partition, YugabyteDBOffsetContext previousOffset)
             throws InterruptedException {
         SnapshottingTask snapshottingTask = getSnapshottingTask(partition, previousOffset);
+        LOGGER.info("Dispatcher in snapshot: " + dispatcher.toString());
```
vrajat, line 107: Better as DEBUG? I am assuming this can happen often

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -166,11 +209,290 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
...
+                  LOGGER.info("Connector has been stopped");
+                  break;
+                }
+                
+                GetChangesResponse resp = this.syncClient.getChangesCDCSDK(table, 
+                    ybConnectorConfig.streamId(), tabletId, cp.getTerm(), cp.getIndex(), cp.getKey(), 
+                    cp.getWrite_id(), cp.getTime(), schemaNeeded.get(tabletId));
+                
+                // Process the response
+                for (CdcService.CDCSDKProtoRecordPB record :
```
vrajat, line 346: You should check for cancel in this loop similar to the streaming change. Otherwise the connector may not be cancellable for a long time.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
vaibhav-yb: There was a `connectorConfig` variable in the parent class as well which was getting initialized in the super constructor. I was just playing around with the name change, have reverted the name change now.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
vaibhav-yb: Done.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
vaibhav-yb: It is already there on line L336

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -192,6 +514,36 @@ protected SnapshotContext<YBPartition, YugabyteDBOffsetContext> prepare(YBPartit
...
+     * @return true if snapshot has been taken or some data has streamed already
+     * @throws Exception if checkpoint cannot be retreived from server side
+     */
+    protected boolean hasSnapshotCompletedPreviously(String tableId, String tabletId) 
+        throws Exception {
+      // The default behaviour is configured to allow the connector to take a snapshot again
+      // and returning false at this stage would indicate that the snapshot has not been taken
+      // so the connector would try to take a snapshot considering this as the first time
+      // it is going for the snapshot
+      if (this.connectorConfig.snapshotAgain()) {
```
vaibhav-yb, line 531: Note to self: handle scenario where all the tablets are in streaming mode and the connector is restarted.

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -1,5 +1,189 @@
...
+                                    + tableInfo.getPgschemaName() + "." 
+                                    + tableInfo.getName();
+              TableId tableId = YugabyteDBSchema.parseWithSchema(fqlTableName, 
+                                                                 tableInfo.getPgschemaName());
+
+              // Retrieve the list of tables in the stream ID,
+              GetDBStreamInfoResponse dbStreamInfoResponse = ybClient.getDBStreamInfo(
+                                                               connectorConfig.streamId());
+
+              if (connectorConfig.getTableFilters().dataCollectionFilter().isIncluded(tableId)
```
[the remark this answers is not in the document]
vaibhav-yb: Yes, for now it will fail. Note that this was the existing behavior, we have just refactored the function to a different utility class.

This behavior will be changed/fixed in the PR which will have the add table support.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -166,11 +209,290 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
...
-      return SnapshotResult.skipped(previousOffset);
+      LOGGER.info("Starting the snapshot process now");
+      
+      // Get the list of tablets
+      List<Pair<String, String>> tableToTabletIds = null;
+      try {
+        String tabletList = this.connectorConfig.getConfig().getString(YugabyteDBConnectorConfig.TABLET_LIST);
+        tableToTabletIds = (List<Pair<String, String>>) ObjectUtil.deserializeObjectFromString(tabletList);
+      } catch (Exception e) {
+        LOGGER.error("The tablet list cannot be deserialized");
```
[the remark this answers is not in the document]
vaibhav-yb: Added the code to throw an exception and fail.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -166,11 +209,290 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
...
+      Map<TableId, String> filteredTableIdToUuid = determineTablesForSnapshot(tableIdToTable);
+
+      Map<String, Boolean> schemaNeeded = new HashMap<>();
+      Set<String> snapshotCompletedTablets = new HashSet<>();
+      Set<String> snapshotCompletedPreviously = new HashSet<>();
+
+      for (Pair<String, String> entry : tableToTabletIds) {
+        schemaNeeded.put(entry.getValue(), Boolean.TRUE);
+
+        previousOffset.initSourceInfo(entry.getValue(), this.connectorConfig);
```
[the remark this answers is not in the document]
vaibhav-yb: Done.

suranjan (approved):

PR #45, opened 2022-09-09
Add PGCompatible Transform

vrajat:
Add a transform to generate the event in PG Compatible schema


vaibhav-yb (approved):
One high level question, if we use this on the source side before publishing the message to Kafka, then I suppose we don't need the transformers on the jdbc-sink side anymore, is my observation correct?

vrajat:
Yes. This transform will be sufficient and YbExtractNewRecordState won't be required on the sink side anymore.

PR #55, opened 2022-10-03
PGCompatible Transform should recurse payload

vrajat:
PGTransform has to transform schemas even if value is NULL. 
PGTransform inherits optional/required attribute correctly.


vaibhav-yb (approved):
The changes seem fine to me. Although I cannot see any GH Action being run on this PR, can you please check it once before merging?

vrajat:
I should enable GH actions for 1.7. I'll do that as well.

PR #56, opened 2022-10-03
PGCompatible Transform should recurse payload

vrajat:
PGTransform has to transform schemas even if value is NULL. 
PGTransform inherits optional/required attribute correctly.

Also enable GitHub Actions on 1.7 branch

### DIFF
--- a/.github/workflows/sanity-workflow.yml
+++ b/.github/workflows/sanity-workflow.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - main
+      - 1.7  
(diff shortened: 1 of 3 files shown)


vaibhav-yb (approved):
Have you tested this transform in an actual Kafka Connect environment? If not, I would want to give it a try once.

Also, can you please add an example in the description - what is the expected output in case this transform is used? It would be helpful in updating our documentation as well.

vrajat:
Yes. Tested for the S3 Blog.

PR #60, opened 2022-10-14
Refactoring of partition classes

vaibhav-yb:
The following changes have been made in this PR:
* Removed class `YugabyteDBPartition` after moving the static `Provider` class to `YBPartition`
* Multiple partitions per tasks are supported now
* Removed a code snippet causing the closure of `YBClient` instance intermittently and causing task failure in tests.


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -603,14 +603,6 @@ private OpId getTransactionStartLsn() throws SQLException {
     @Override
     protected void complete(SnapshotContext<YBPartition, YugabyteDBOffsetContext> snapshotContext) {
         snapshotter.snapshotCompleted();
-
```
vrajat, line 606: Can you help me understand this removal? The connections should not be closed after snapshot because the same connections are used in streaming?

vrajat (commented):
I had made similar changes to YBPartition and offsets and I had broken the connector. I had to revert these changes. I think you had found the root cause (thanks again! :)). What is different this time that the getPreviousOffsets fn will work.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
vaibhav-yb: This removal is more from the test POV, some of the tests seemed to be failing intermittently saying that the client connection was closed in this function.

Then on investigation, I realised this `complete()` function is to just mark the snapshot as completed, the closing of `YBClient` instances should be done at the very end of the connector i.e. when it is killed or stopped - and we have a function doing the same already in `YugabyteDBConnector` so the reason why I was getting the exception was that I was simply trying to stop a already stopped instance of the client instance.

vrajat (approved):

PR #61, opened 2022-10-14
Changes to support tablet split

vaibhav-yb:
With the merge of this PR, the connector will start supporting tablet splitting on the YugabyteDB server side. Once a tablet split is detected, the connector will gracefully handle the deletion of the parent tablet from its cache and it will add the child tablets to its cache so that they can be polled further for the data.

Also, added a unit test i.e. `YugabyteDBTabletSplitTest` to verify the tablet split connector behavior.


vaibhav-yb:
This PR has a dependency on PR #60 so please merge that one before proceeding with this PR.

pom.xml
```diff
@@ -46,7 +46,8 @@
         <version.kafka>3.2.1</version.kafka>
         <version.org.slf4j>1.7.36</version.org.slf4j>
         <version.logback>1.4.0</version.logback>
-        <version.ybclient>0.8.21-20220715.193438-1</version.ybclient>
+        <!-- <version.ybclient>0.8.21-20220715.193438-1</version.ybclient> -->
+        <version.ybclient>0.8.28-SNAPSHOT</version.ybclient>
```
vrajat, line 50: Did you compile locally? The version should point to one with timestamp and not SNAPSHOT.

src/main/java/io/debezium/connector/yugabytedb/SourceInfo.java
```diff
@@ -95,11 +95,19 @@ public final class SourceInfo extends BaseSourceInfo {
     private String tableName;
     private String tabletId;
 
+    // todo Vaibhav: maybe add another constructor in order to initialize the SourceContext properly
     protected SourceInfo(YugabyteDBConnectorConfig connectorConfig) {
         super(connectorConfig);
         this.dbName = connectorConfig.databaseName();
     }
 
+    protected SourceInfo(YugabyteDBConnectorConfig connectorConfig, OpId lastCommitLsn) {
```
vrajat, line 104: what will be the lsn if the original constructor is used?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBErrorHandler.java
```diff
@@ -23,6 +24,15 @@ public YugabyteDBErrorHandler(YugabyteDBConnectorConfig connectorConfig, ChangeE
 
     @Override
     protected boolean isRetriable(Throwable throwable) {
+        // Error throwable class indicating tablet split
```
vrajat, line 27: For my understanding, is this what triggers a restart?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -484,8 +489,11 @@ else if (message.isDDLMessage()) {
                     // has succeeded
                     retryCount = 0;
                 }
-            }
-            catch (Exception e) {
+            } catch (CDCErrorException cdcException) {
+                // This will cause the connector to restart
+                LOGGER.info("Encountered CDCErrorException, possibly indicating a tablet split", cdcException);
```
vrajat, line 494: Why guess? Is it possible to have a specific exception for tablet split?

pom.xml
vaibhav-yb: Yes, because at the time of compilation, the upstream one was not present, will change it now.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBErrorHandler.java
vaibhav-yb: Yes, this restarts the task. But we are not doing the connector restart now, the tablet split is handled gracefully without a restart.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
vaibhav-yb: Removed this log line, added it for testing before.

src/main/java/io/debezium/connector/yugabytedb/SourceInfo.java
vaibhav-yb: It will be null, but that constructor is only used at one place while starting the connector for the first time, after which the connector calls the function `SourceInfo#update()` to change all those values.

Also yes, I believe you're trying to indicate a potential `NullPointerException` if this is not coded properly, but this is again handled in the class `YugabyteDBOffsetContext`, see the below code snippet:
```java
OpId lsn() {
    return sourceInfo.lsn() == null ? new OpId(0, 0, null, 0, 0)
        : sourceInfo.lsn();
}
```

So in case the original constructor is used, we will try to simply ask the server to send the changes from the beginning, but the server will only send from the checkpoint which has not been read, and then when we get those new changes, we can simply update the lsn for the tablet.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBErrorHandler.java
```diff
@@ -6,6 +6,7 @@
 package io.debezium.connector.yugabytedb;
 
 import org.postgresql.util.PSQLException;
+import org.yb.client.CDCErrorException;
```
vrajat, line 9: Can this import be removed now?

vrajat (commented):
Where will the tests be added for this? Integration tests in `cdcsdk-test`?

vrajat (approved):
Very nice test!

PR #62, opened 2022-10-17
Changes to support dynamic addition of new tables

vaibhav-yb:
* Added a new poller thread to check if there are any new tables associated with the provided stream ID
* Added a configuration parameter - `auto.add.new.tables` with default: `true` in case a user doesn't want to dynamically add the tables to the connector


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -286,6 +286,14 @@ private void getChanges2(ChangeEventSourceContext context,
             bootstrapTabletWithRetry(tabletPairList);
         }
 
+        // Launch a thread to poll for the tables.
```
vrajat, line 289: An executor service should be used instead. Specifically a fixed thread pool. Also code to shutdown the thread is required.

vrajat: On more thoughts a [scheduled executor service](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledExecutorService.html) is better

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBErrorHandler.java
```diff
@@ -30,6 +33,13 @@ protected boolean isRetriable(Throwable throwable) {
             return true;
         }
 
+        // This condition will be hit when the poller thread indicates that a new table has been
+        // found to be associated with the provided stream ID.
+        if (throwable instanceof NewTableFoundException) {
```
vrajat, line 38: Don't understand the workflow. Thread throws an exception and this part gets called? Is that why you use thread instead of executor service?

vrajat (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
vaibhav-yb: Currently, I followed the way this [TablePollerThread](https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/main/java/io/confluent/connect/jdbc/source/TableMonitorThread.java) is implemented for the `kafka-connect-jdbc`.

I can still implement the `ScheduledExecutorService` changes if required.

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -12,7 +12,11 @@
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+<<<<<<< HEAD
```
vrajat, line 15: Looks like a merge conflict has sneaked in.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnector.java
```diff
@@ -196,6 +205,14 @@ private YBClient getYBClientBase(String hostAddress, long adminTimeout, long ope
 
     @Override
     public void stop() {
+        LOGGER.info("Stopping table monitoring thread");
+        tableMonitorThread.shutdown();
+        try {
+            tableMonitorThread.join(MAX_TIMEOUT);
+        } catch (InterruptedException ie) {
+            // Ignore, shouldn't be interrupted.
```
vrajat, line 213: Can you put a warning message just as a fall back that it may not have worked.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBTablePoller.java
```diff
@@ -0,0 +1,138 @@
...
+
+  private final YugabyteDBConnectorConfig connectorConfig;
+  private final ConnectorContext connectorContext;
+  private final YBClient ybClient;
+  private final CountDownLatch shutdownLatch;
+  private final long pollMs;
+
+  private short retryCount = 0;
+  
+  private List<TableInfo> cachedTableInfoList = null;
```
vrajat, line 33: Is Set a better option? In kafka-connect-jdbc, they are figuring out duplicates etc which is not here.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBTablePoller.java
```diff
@@ -0,0 +1,138 @@
...
+        LOGGER.debug("Cached table list in the poller thread is null, initializing it now");
+        cachedTableInfoList = resp.getTableInfoList();
+      } else {
+        if (cachedTableInfoList.size() != resp.getTableInfoList().size()) {
+          // TODO: We can also check for a condition whether the new added tables are a part of the
+          // include list (if they satisfy the regex criteria), if they don't satisfy then we
+          // should not restart.
+
+          String message = "Found {} new table(s), signalling context reconfiguration";
+          LOGGER.info(message, resp.getTableInfoList().size() - cachedTableInfoList.size());
```
vrajat, line 88: If you decide to use Set instead of List, can you print intersection & difference if debug logging is turned on? 
Refer: https://stackoverflow.com/a/3590718

vrajat (changes requested):
One minor issue with conflicts. That you have to change. I'll resolve the executor service comment. 

Can you look into using set instead of list?

Also which of the tests are applicable here? https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/test/java/io/confluent/connect/jdbc/source/TableMonitorThreadTest.java

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBTablePoller.java
vaibhav-yb: Added the code to print intersection and difference while debug logging is on.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBTablePoller.java
vaibhav-yb: Yes, set seems to be a better option. Changing the list to a set. Some future work is planned to also verify the table names i.e. if a new table is added then we should also verify if it is a part of the include list.

Currently, we are slowed down by a limitation on the server side that the pgschema name is only available in a certain ListTablesResponse object, to use the same in this piece of code would make the code a little bit too complicated. I was thinking of implementing those things in further releases.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnector.java
vaibhav-yb: Done.

vaibhav-yb:
@vrajat 
> Also which of the tests are applicable here? https://github.com/confluentinc/kafka-connect-jdbc/blob/master/src/test/java/io/confluent/connect/jdbc/source/TableMonitorThreadTest.java

Some of these tests are applicable here but adding them would require infra changes now, I am not sure if we want to do that immediately. On the other hand, we can add tests in the `cdcsdk-testing` easily.

vrajat (approved):

PR #65, opened 2022-11-04
Addition of metric classes to support multi partition per task metrics

vaibhav-yb:
This PR contains the diff to have the metric classes managed by YugabyteDB maintainers. Additional changes are made there to support the multi-partition per task model the Debezium Connector for YugabyteDB follows.


src/main/java/io/debezium/connector/yugabytedb/metrics/YugabyteDBStreamingPartitionMetrics.java
```diff
@@ -0,0 +1,70 @@
...
+public class YugabyteDBStreamingPartitionMetrics extends AbstractYugabyteDBPartitionMetrics
+        implements YugabyteDBStreamingPartitionMetricsMXBean {
+
+    private final StreamingMeter streamingMeter;
+
+    public YugabyteDBStreamingPartitionMetrics(CdcSourceTaskContext taskContext,
+                                               Map<String, String> tags,
+                                               EventMetadataProvider metadataProvider) {
+        super(taskContext, tags, metadataProvider);
+        streamingMeter = new StreamingMeter(taskContext, metadataProvider);
```
vrajat, line 31: Where are these metrics set?

vaibhav-yb: The streaming meter metrics are set by some `onEvent()` function which Debezium has implemented in its core: https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/pipeline/meters/StreamingMeter.java#L69

There are other meters as well: https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/pipeline/meters/

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -272,11 +272,6 @@ private void getChanges2(ChangeEventSourceContext context,
             schemaNeeded.put(entry.getValue(), Boolean.TRUE);
         }
 
-        for (Pair<String, String> entry : tabletPairList) {
-            final String tabletId = entry.getValue();
-            offsetContext.initSourceInfo(tabletId, this.connectorConfig);
-        }
-
```
vaibhav-yb, line 279: Removed this block since it is overriding what was being initialized in the previous block.

Ideally this should not cause any issues since we call GetChanges from 0,0 and get from certain OpId and then update the offset.

vrajat (approved):
Approving since #71 will track regressions in this PR

Issue #71, opened 2022-11-14
Metrics not getting published at the task level


fourpointfour:
With the recent changes to the metrics classes, the issue where we were failing to register metrics is gone and we can now register metrics after making changes where we are handling multiple partitions per task.

However, with this, an issue has been observed that the metrics for all the tasks are not published on the endpoint, this needs an investigation and a fix.

PR #72, opened 2022-11-15
Properly publish metrics at the task level for each partition

vaibhav-yb:
This PR contains the diff for the following:
1. Fix for some debezium metrics not getting published - the issue was with the wrong order of tags in the name of the metrics, the order stands corrected now.
2. Removal of a class i.e. `YugabyteDBMetrics`

Additionally, this will close issue #71


vaibhav-yb:
Abandoning this PR since it is not required now. The changes, as required were needed in [yugabyte/cdc-examples](https://github.com/yugabyte/cdc-examples) repository and they have been taken care of now.

Issue #73, opened 2022-11-16
GitHub actions get stuck on pipeline runs


vaibhav-yb:
After merging the recent changes, the GitHub Actions and the Jenkins pipeline running the tests either fail or get stuck at a certain point because they are unable to fetch the changes or compile properly. This is because the Docker image we are specifying to run the tests against doesn't contain the required server side changes to run some tests and that's why the tests either get stuck waiting for change records or fail.

**Suggested fix:**
Pull the docker image from a repository on Quay before running GH actions so that we always have the latest image available with us.

PR #74, opened 2022-11-16
[#73] GitHub actions stuck while running the tests

vaibhav-yb:
This PR contains the diff to use a docker image as specified in the environment variable. The variable picks up the version from a stored GitHub action secret.

Additionally, it contains two more fixes to the test infrastructure:
1. Fail the tests if they are stuck for more than 5 minutes
2. Fix for a failing test because incorrect insert string was being used

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -604,22 +604,8 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
     protected void complete(SnapshotContext<YBPartition, YugabyteDBOffsetContext> snapshotContext) {
         snapshotter.snapshotCompleted();
 
-        // Close the YBClient instances now
-        if (asyncClient != null) {
-          try {
-            asyncClient.close();
-          } catch (Exception e) {
-            LOGGER.warn("Error while closing async YBClient", e);
-          }
-        }
-
-        if (syncClient != null) {
-          try {
-            syncClient.close();
-          } catch (Exception e) {
-            LOGGER.warn("Error while closing YBClient", e);
-          }
-        }
+        // Todo Vaibhav: Close the YBClient instances now
+        // See if it can be closed anywhere else for the snapshotting tasks.
     }
 
     /**
--- a/src/test/java/io/debezium/connector/yugabytedb/TestHelper.java
+++ b/src/test/java/io/debezium/connector/yugabytedb/TestHelper.java
@@ -376,7 +376,17 @@ public final class TestHelper {
     }
 
     protected static YugabyteYSQLContainer getYbContainer() {
-        YugabyteYSQLContainer container = new YugabyteYSQLContainer("yugabytedb/yugabyte:2.17.1.0-b128");
+        String dockerImageName = System.getenv("YB_DOCKER_IMAGE");
+        
+        if (dockerImageName == null || dockerImageName.isEmpty()) {
+            dockerImageName = "yugabytedb/yugabyte:latest";
+        }
+
+        LOGGER.info("Using docker image in test: {}", dockerImageName);
+        
+        YugabyteYSQLContainer container = new YugabyteYSQLContainer(
+            DockerImageName.parse(dockerImageName)
+            .asCompatibleSubstituteFor("yugabytedb/yugabyte"));
         container.withPassword("yugabyte");
         container.withUsername("yugabyte");
         container.withDatabaseName("yugabyte");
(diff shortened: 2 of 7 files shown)


src/test/java/io/debezium/connector/yugabytedb/TestHelper.java
```diff
@@ -566,6 +581,23 @@ protected static void waitFor(Duration duration) {
             });
     }
 
+
+    /**
+     * Fail the test if the time exceeds more than 5 minutes to wait for the values.
+     * @param startTime the start time of the verification in milliseconds
+     * @param secondsToWait total duration to wait before failing the test
+     */
+    protected static void failTestIfTimeExceeds(long startTime, long secondsToWait) {
```
vrajat, line 590: This function can be replaced with [awaitility + lambdas](). 


For example:

    await().atLeast(1, SECONDS).and().atMost(2, SECONDS).until(value(), equalTo(1));
    
This means that the loops in which it is used will also change. For example:

    await().atmost(10, seconds).until( () -> {
       int consumed = super.consumeAvailableRecords(record -> {
           LOGGER.debug("The record being consumed is " + record);
           records.add(record);
        });
        if (consumed > 0) {
            totalConsumedRecords += consumed;
            LOGGER.debug("Consumed " + totalConsumedRecords + " records");
        }
        return totalConsumedRecords == recordsCount;
     });


If lambda doesn't wok, you can move it into a fn. as well.
Also a polling interval is a good idea.

vrajat (changes requested):
Is a polling interval also required in these tests?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -604,22 +604,8 @@ private OpId getTransactionStartLsn() throws SQLException {
...
-
-        if (syncClient != null) {
-          try {
-            syncClient.close();
-          } catch (Exception e) {
-            LOGGER.warn("Error while closing YBClient", e);
-          }
-        }
+        // Todo Vaibhav: Close the YBClient instances now
+        // See if it can be closed anywhere else for the snapshotting tasks.
```
vaibhav-yb, line 608: I have removed the closure of clients because it seems the `execute()` function is being called per partition and once it finishes, it closes the YBClient instance, which are then rendered unusable later on for the same task so it started giving exceptions that the yb-client object has already been closed.

Currently I am relying on the garbage-collection mechanism that it will collect this open object, but I will soon raise another PR to fix this specific issue.

vrajat: Can you also create a issue?

src/test/java/io/debezium/connector/yugabytedb/TestHelper.java
```diff
@@ -426,7 +436,12 @@ public static String getNewDbStreamId(String namespaceName, String tableName) th
             throw new NullPointerException("No table found with the name " + tableName);
         }
 
-        return syncClient.createCDCStream(placeholderTable, namespaceName, "PROTO", "IMPLICIT").getStreamId();
+        String dbStreamId = syncClient.createCDCStream(placeholderTable, namespaceName,
```
vrajat, line 439: Can this statement throw an exception? I suggest wrapping it `try ... finally`

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
```diff
@@ -141,21 +135,46 @@ private void verifyDeletedFieldPresentInValue(long recordsCount, YBExtractNewRec
...
+     * @param recordsCount total number of records which should be consumed
+     * @param milliSecondsToWait duration in milliseconds to wait for while consuming
+     */
+    private void waitAndFailIfCannotConsume(List<SourceRecord> records, long recordsCount,
+                                            long milliSecondsToWait) {
+        AtomicLong totalConsumedRecords = new AtomicLong();
+        long seconds = milliSecondsToWait / 1000;
+        try {
+            Awaitility.await()
+                .atMost(Duration.ofSeconds(seconds))
```
vrajat, line 151: Is a polling interval also required?

vrajat (commented):
I have a few minor comments. Otherwise looks good. Will approve after I get a response from you.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
vaibhav-yb: Created issue #80

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
vaibhav-yb: We can add it, but it may unnecessarily delay the test execution time.

src/test/java/io/debezium/connector/yugabytedb/TestHelper.java
vaibhav-yb: Yes, this statement can throw an exception. However, the function itself checks this exception:
```
public static String getNewDbStreamId(String namespaceName, String tableName) throws Exception {
...
}
```

Is a `try...finally` block still needed?

vrajat: Then the client won't be closed.

vrajat (approved):

PR #84, opened 2022-11-18
[#82] Support for publishing before image of the row

### LINKED ISSUE #82: Add before image support
As per current status quo, the connector will not be able to send the before image of the updated row.

The issue requires fixing/addition of the support so that the connector can send the before image of the row once it is updated.

vaibhav-yb:
The changes introduced in this PR are all towards processing the before image of the row. Before this PR, when encountered an `UPDATE` op, and the function `getOldColumnValues()` was being called, we simply returned a `null` value assuming that since we are not even publishing the old value in our change record from server side, why even bother processing it.

With the latest changes on server side made to enable old image / before image of the changed row, we now publish them with the change records from the server and in that case, when we call `getOldColumnValues()`, the function should be able to process the messages properly in case of an update operation and hence the change in the mentioned function.

Now, as an optimization for the default cases, where the stream ID is not configured to stream before images, we have added a boolean which will be set while streaming to tell the change record emitter whether it needs to process the records for before image or not.

Additionally, this PR closes #82

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
@@ -13,9 +13,12 @@ import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.yb.client.AsyncYBClient;
+import org.yb.client.CDCStreamInfo;
+import org.yb.cdc.CdcService.CDCRecordType;
 import org.yb.cdc.CdcService.TabletCheckpointPair;
 import org.yb.client.GetDBStreamInfoResponse;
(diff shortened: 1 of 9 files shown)


src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -229,4 +232,54 @@ public static OpId getOpIdFromGetTabletListResponse(GetTabletListToPollForCDCRes
     // Return null if no match is found, ideally this shouldn't happen in any case
     return null;
   }
+
+  /**
+   * Check whether the passed stream ID in the connector configuration has before image enabled.
```
suranjan, line 237: We should do this in the YB*COnnector and send that as a boolean prop to each task.

suranjan (changes requested):

suranjan (approved):

PR #89, opened 2022-11-21
[#88] Added logging around potential null pointers

### LINKED ISSUE #88: Connector throws `NullPointerException` when the message in exception is empty
In integration test runs, it has been observed a few times, that an exception is thrown which has no error message and this may lead to a `NullPointerException` in `YugabyteDBErrorHandler`.

It can be prevented by adding a check in `YugabyteDBErrorHandler` to not proceed ahead if the error message is found to be null/empty.

vaibhav-yb:
This PR closes #88

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBErrorHandler.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBErrorHandler.java
@@ -6,6 +6,8 @@
 package io.debezium.connector.yugabytedb;
 
 import org.postgresql.util.PSQLException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import io.debezium.connector.base.ChangeEventQueue;
 import io.debezium.pipeline.ErrorHandler;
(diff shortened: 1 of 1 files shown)


suranjan (commented):
I would like to see in which case we get the msg null for debugging purposes. Can we print that exception as well?

vaibhav-yb:
> I would like to see in which case we get the msg null for debugging purposes. Can we print that exception as well?

@suranjan I have added an info log which will print the exception in all the cases so whenever we get this null log, we can simply see the log line just before that and analyse it for debugging purposes.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBErrorHandler.java
```diff
@@ -16,14 +18,22 @@
...
     public YugabyteDBErrorHandler(YugabyteDBConnectorConfig connectorConfig, ChangeEventQueue<?> queue) {
         super(YugabyteDBConnector.class, connectorConfig, queue);
     }
 
     @Override
     protected boolean isRetriable(Throwable throwable) {
+        LOGGER.info("Checking if a retry is possible for retriable exception: {}", throwable);
+
+        if (throwable.getMessage() == null) {
+            LOGGER.warn("Exception message received is null");
```
suranjan, line 32: We should log the throwable if it is not being logged.

suranjan (changes requested):

vrajat (approved):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBErrorHandler.java
vaibhav-yb: It is being logged in line number 29.

PR #107, opened 2022-12-02
Kill all tasks when we catch InterruptedException, minor logging changes

adithya-yb:
Previously, when we caught a  InterruptedException exception in the method: getChanges2
We were simply printing the stack trace and moving on. Now if one of the task in the execution service throws an 'InterruptedException' , we will kill all the tasks. 

Also adding minor logging changes: 
1. Every log from a task will also pint the task_id as part of the context
2. When we encounter an exception through 'GetChanges', we will also print the tablet causing the exception

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
@@ -144,7 +144,7 @@ public class YugabyteDBConnectorTask
                                                                       connectorConfig);
 
         LoggingContext.PreviousContext previousContext = taskContext
-                .configureLoggingContext(CONTEXT_NAME + "_" + taskId);
+                .configureLoggingContext(CONTEXT_NAME + "|" + taskId);
         try {
             // Print out the server information
             // CDCSDK Get the table,
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
@@ -134,17 +134,17 @@ public class YugabyteDBConnectorTask
         this.taskContext = new YugabyteDBTaskContext(connectorConfig, schema, topicSelector, taskId);
 
         // Get the tablet ids and load the offsets
-        final Offsets<YBPartition, YugabyteDBOffsetContext> previousOffsets = 
+        final Offsets<YBPartition, YugabyteDBOffsetContext> previousOffsets =
             getPreviousOffsetsFromProviderAndLoader(
                 new YBPartition.Provider(connectorConfig),
                 new YugabyteDBOffsetContext.Loader(connectorConfig));
         final Clock clock = Clock.system();
 
-        YugabyteDBOffsetContext context = new YugabyteDBOffsetContext(previousOffsets, 
+        YugabyteDBOffsetContext context = new YugabyteDBOffsetContext(previousOffsets,
                                                                       connectorConfig);
 
         LoggingContext.PreviousContext previousContext = taskContext
-                .configureLoggingContext(CONTEXT_NAME);
+                .configureLoggingContext(CONTEXT_NAME + "_" + taskId);
         try {
             // Print out the server information
             // CDCSDK Get the table,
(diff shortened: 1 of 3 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
```diff
@@ -134,17 +134,17 @@ public ChangeEventSourceCoordinator<YBPartition, YugabyteDBOffsetContext> start(
...
                 new YugabyteDBOffsetContext.Loader(connectorConfig));
         final Clock clock = Clock.system();
 
-        YugabyteDBOffsetContext context = new YugabyteDBOffsetContext(previousOffsets, 
+        YugabyteDBOffsetContext context = new YugabyteDBOffsetContext(previousOffsets,
                                                                       connectorConfig);
 
         LoggingContext.PreviousContext previousContext = taskContext
-                .configureLoggingContext(CONTEXT_NAME);
+                .configureLoggingContext(CONTEXT_NAME + "_" + taskId);
```
vaibhav-yb, line 147: This may change the way we publish metrics as well maybe since we use the task context as one of the keys in forming the name of the metric.

vaibhav-yb (commented):
Will it be possible to remove these refactoring changes from this PR? Not an issue though, we can have it like this.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
vaibhav-yb: But yes, on another thought, the way we form a key is in the pattern:
```
context=streaming,taskId=0
```
and after your change, it might get published as
```
context=streaming_0,taskId=0
```

And the way we detect this metrics is by providing regex like (asterisk is for reference only):
```
context=* taskId=*
```
so ideally we will not be breaking anything.

vrajat:
@adithya-yb is there an issue or a slack conversation for why this change is required? Curios what scenario motivates this change.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -494,8 +496,7 @@ else if (message.isDDLMessage()) {
                                 }
                             } catch (InterruptedException ie) {
                                 ie.printStackTrace();
-                            } catch (SQLException se) {
```
vrajat, line 497: Is the sqlexception catch removed? Can you explain why?

vrajat:
@vaibhav-yb are the checks failing because this branch does not have your changes from main?

vaibhav-yb:
> @vaibhav-yb are the checks failing because this branch does not have your changes from main?

Yes @vrajat I have merged my changes to this branch now, actions should run.

vrajat:
There are a couple of failures:

```
YugabyteDBColocatedTablesTest.shouldWorkWithMixOfColocatedAndNonColocatedTables:155 NullPointer
[22671](https://github.com/yugabyte/debezium-connector-yugabytedb/pull/107/checks?notification_referrer_id=NT_kwDOABj_mrI0OTgxMjkyMzAwOjE2MzgyOTg#step:6:22672)
Error:    YugabyteDBDatatypesTest.testEnumValue:410 » Execution java.lang.RuntimeException: java.util.concurrent.CompletionException: org.opentest4j.AssertionFailedError: Failed to consume 3 in 300 seconds
```

Expected?

vaibhav-yb:
> There are a couple of failures:
> 
> ```
> YugabyteDBColocatedTablesTest.shouldWorkWithMixOfColocatedAndNonColocatedTables:155 NullPointer
> [22671](https://github.com/yugabyte/debezium-connector-yugabytedb/pull/107/checks?notification_referrer_id=NT_kwDOABj_mrI0OTgxMjkyMzAwOjE2MzgyOTg#step:6:22672)
> Error:    YugabyteDBDatatypesTest.testEnumValue:410 » Execution java.lang.RuntimeException: java.util.concurrent.CompletionException: org.opentest4j.AssertionFailedError: Failed to consume 3 in 300 seconds
> ```
> 
> Expected?

Yes @vrajat , these are the intermittent failures we were observing. Something seems to be off when they are run with the complete test suite, when run individually they both are passing.

vaibhav-yb (approved):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
adithya-yb: This is no longer needed since we will not throw a 'SQLException' from any part of the codec within the try.

adithya-yb:
#111 Kill all tasks when we catch InterruptedException, minor logging changes

PR #113, opened 2022-12-09
Log GetChanges request at 5 minutes intervals

vaibhav-yb:
This PR adds the change so that the connector also logs the `GetChanges` requests at regular intervals so that we know that the tasks are active and requesting for changes actively.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -322,6 +322,10 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
       }
 
       short retryCount = 0;
+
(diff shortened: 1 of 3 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -366,8 +370,11 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
 
                 OpId cp = previousOffset.snapshotLSN(tabletId);
 
-                LOGGER.debug("Going to fetch from checkpoint {} for tablet {} for table {}", 
-                            cp, tabletId, table.getName());
+                if (LOGGER.isDebugEnabled() || System.currentTimeMillis() >= (lastLoggedTimeForGetChanges + 300_000)) {
```
suranjan, line 373: Always use a config for these so that we can enable/disable. Please add a config.

suranjan (approved):

PR #121, opened 2022-12-12
Kill complete connector if one task fails

### LINKED ISSUE #120: Kill the connector if one of the task fails
At the current moment, we throw the exception as we receive it in case all the 5 retries of a connector fails. However, in case the flow goes to the `ErrorHandler` class, we simply kill the current task and go ahead with the other tasks - this may lead to data loss in certain scenarios where a task stops streaming data.

The objective of this issue is to make the connector fail as soon as some task fails so that the errors can be detected early as soon as the connector stops so that the user takes necessary action at that point in time itself.

vaibhav-yb:
This closes #120 by killing the complete connector in case even a single task fails, streaming will be stopped after that.


suranjan (commented):
Have we tested this?

PR #125, opened 2022-12-21
Request schema from the server once we start polling for child tablets

vaibhav-yb:
After a tablet has been split, we need to explicitly mention the flag in the GetChanges call so that the server sends the schema for the child tablet and we register it for further processing. This PR implements the same.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSchema.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSchema.java
@@ -120,7 +120,7 @@ public class YugabyteDBSchema extends RelationalDatabaseSchema {
                                              String schemaName,
                                              String tabletId) {
         String lookupKey = getLookupKey(tableId, tabletId);
-        if (cdcsdkSchemaPB == null) {
+        if (tabletIdToCdcsdkSchemaPB.get(lookupKey) == null || cdcsdkSchemaPB == null) {
             tabletIdToCdcsdkSchemaPB.put(lookupKey, schemaPB);
         }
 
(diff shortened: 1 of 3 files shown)


vrajat (approved):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -751,8 +756,8 @@ private void addTabletIfNotPresent(List<Pair<String,String>> tabletPairList,
 
             LOGGER.info("Initialized offset context for tablet {} with OpId {}", tabletId, OpId.from(pair.getCdcSdkCheckpoint()));
 
-            // Add the flag to indicate that we do not need the schema from this tablet again.
-            schemaNeeded.put(tabletId, Boolean.FALSE);
+            // Add the flag to indicate that we need the schema for the new tablets so that the schema can be registered.
+            schemaNeeded.put(tabletId, Boolean.TRUE);
```
suranjan, line 760: have we checked that we don't fetch schema everytime?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSchema.java
```diff
@@ -120,7 +120,7 @@ protected void refreshSchemaWithTabletId(TableId tableId,
                                              String schemaName,
                                              String tabletId) {
         String lookupKey = getLookupKey(tableId, tabletId);
-        if (cdcsdkSchemaPB == null) {
+        if (tabletIdToCdcsdkSchemaPB.get(lookupKey) == null || cdcsdkSchemaPB == null) {
```
suranjan, line 123: containsKey is better, can we have null value against the key?

vaibhav-yb: Addressed.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
vaibhav-yb: Yes, as soon as we get a DDL, we set this flag to false while processing the records in the for loop.

PR #130, opened 2022-12-26
Stream consistency bug fixes and test additions

vaibhav-yb:


vrajat (approved):

vrajat (approved):
LGTM up to this point.

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
```diff
@@ -4,22 +4,29 @@
...
 
+/**
+ * @author Rajat Venkatesh
+ */
 public class Merger {
     private static final Logger LOGGER = LoggerFactory.getLogger(Merger.class);
-    private final PriorityQueue<Message> queue = new PriorityQueue<>();
-    private final Map<String, List<Message>> mergeSlots = new HashMap<>();
-    private final Map<String, Long> tabletSafeTime = new HashMap<>();
+    private final PriorityBlockingQueue<Message> queue = new PriorityBlockingQueue<>();
```
vrajat, line 18: Can you move the data structures back to non-concurrent ones? There is no point in taking locks at function and data structure level.

vrajat (changes requested):

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
```diff
@@ -31,32 +35,51 @@ public void addMessage(Message message) {
...
         return peeked;
     }
 
-    public Message poll() {
+    public synchronized Optional<Message> poll() {
+        if (queue.isEmpty()) {
+            return Optional.empty();
+        }
+
         Message message = Objects.requireNonNull(queue.poll());
```
vrajat, line 72: `poll` removes the message. However, the next if condition maybe false and might lose the message. Instead the code should `peek` instead. 

My suggestion is turn `Merger::peek` to private and then use it in `Merger::poll`. So the code structure is as follows:

```
public synchronised Optional<Message> poll() {
    Message message = this.peek();
    // No valid message to emit
    if (message.isEmpty())
        return message;
        
    // Remove message from queue and mergeSlots
    queue.poll();
    mergeSlots.get(...).removeIf(...);
    return message;    
}
```

vrajat (changes requested):

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
vaibhav-yb: Addressed.

vrajat (approved):
One minor comment. A lot of the log messages are INFO. Do you want to move it to DEBUG?

vaibhav-yb:
> One minor comment. A lot of the log messages are INFO. Do you want to move it to DEBUG?

@vrajat Wanted to keep the info logs for now as they will help in further debugging in our tests as of now. I will make them as DEBUG once a PR will be opened to merge to `main`.

PR #132, opened 2022-12-28
Change Dockerfile to download the required dependencies

### LINKED ISSUE #131: Remove dependency of Dockerfile on the custom base image
The base image we are using in the `Dockerfile` is a little old now and we need to update it to the latest we have on quay: `quay.io/yugabyte/connect-base-yb:0.4`

vaibhav-yb:
Before this, we used a base image with all the required dependencies (jars) which can then be used further by the image. This PR modifies the Dockerfile so that the process now downloads the jar files directly to the image while building. This PR closes #131

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,11 +3,11 @@
 
 # The base image is derived from debezium/connect:1.7 and contains the following drivers
 # and connectors prepopulated:
-#  - JDBC Sink Connector v10.2.5
+#  - JDBC Sink Connector v10.6.0
 #  - YugabyteDB JDBC Driver v42.3.5-yb-1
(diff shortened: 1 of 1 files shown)


vrajat:
Is it possible to use download these dependencies using `curl` just like imx exporter? This way we don't have to depend on an unknown process to update the image. I can help. How to do you manually download the jars?

vaibhav-yb:
> Is it possible to use download these dependencies using `curl` just like imx exporter? This way we don't have to depend on an unknown process to update the image. I can help. How to do you manually download the jars?

Yes @vrajat it is possible since I download the dependencies using wget even today and then create the base image.

vrajat (approved):

PR #135, opened 2022-12-29
Retry mechanism around the call to `GetTabletListToPollForCDC` while handling tablet split

### LINKED ISSUE #133: Add retry around `GetTabletListToPollForCDC`
Currently, after we encounter a tablet split, we simply head over to handling the split in a function `handleTabletSplit()` - the function calls the api `YBClient#getTabletListToPollForCDC` so as to get a child tablet for the given split parent.

However, there are chances where the API can fail or throw exceptions. The goal is to add a retry around the API call itself so that at least the connector retries appropriately before failing the execution.

vaibhav-yb:
This PR closes #133

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
@@ -357,7 +357,7 @@ public class YugabyteDBStreamingChangeEventSource implements
                       } catch (CDCErrorException cdcException) {
                         // Check if exception indicates a tablet split.
                         LOGGER.info("Code received in CDCErrorException: {}", cdcException.getCDCError().getCode());
-                        if (cdcException.getCDCError().getCode() == Code.TABLET_SPLIT) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -814,4 +806,67 @@ private void handleTabletSplit(CDCErrorException cdcErrorException,
...
+                    throw new Exception("Found unexpected ("
+                            + response.getTabletCheckpointPairListSize()
+                            + ") number of tablets while trying to fetch the children of parent "
+                            + splitTabletId);
+                }
+
+                retryCount = 0;
+                return response;
+            } catch (Exception e) {
+                ++retryCount;
```
vrajat, line 846: Should this code check for an exception type for retry or always retry>

vaibhav-yb: We have noticed during nemesis tests that the exceptions thrown are not specific, sometimes it's `NonRecoverableException`, sometimes `NotFound` so that's why we wanted to retry always.

vrajat (approved):

PR #142, opened 2023-01-18
[CDCSDK] Remove dependency on JDBC connection for decoding values

fourpointfour:
This PR has the changes for:
1. Removing the dependency on JDBC connection while decoding values
2. Modifying a test with the values for arrays as well

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeRecordEmitter.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeRecordEmitter.java
@@ -42,7 +42,6 @@ public class YugabyteDBChangeRecordEmitter extends RelationalChangeRecordEmitter
     private final ReplicationMessage message;
     private final YugabyteDBSchema schema;
     private final YugabyteDBConnectorConfig connectorConfig;
-    private final YugabyteDBConnection connection;
     private final TableId tableId;
 
     private boolean shouldSendBeforeImage = false;
@@ -52,7 +51,7 @@ public class YugabyteDBChangeRecordEmitter extends RelationalChangeRecordEmitter
 
     public YugabyteDBChangeRecordEmitter(YBPartition partition, OffsetContext offset, Clock clock,
                                          YugabyteDBConnectorConfig connectorConfig,
-                                         YugabyteDBSchema schema, YugabyteDBConnection connection,
+                                         YugabyteDBSchema schema,
                                          TableId tableId, ReplicationMessage message, String pgSchemaName,
                                          String tabletId, boolean shouldSendBeforeImage) {
         super(partition, offset, clock);
@@ -60,7 +59,6 @@ public class YugabyteDBChangeRecordEmitter extends RelationalChangeRecordEmitter
         this.schema = schema;
         this.message = message;
         this.connectorConfig = connectorConfig;
-        this.connection = connection;
 
         this.pgSchemaName = pgSchemaName;
 
@@ -184,16 +182,13 @@ public class YugabyteDBChangeRecordEmitter extends RelationalChangeRecordEmitter
                 ? schemaColumns.size()
                 : columnsWithoutToasted.size()];
 
-        final Set<String> undeliveredToastableColumns = new HashSet<>(schema
-                .getToastableColumnsForTableId(table.id()));
         for (ReplicationMessage.Column column : columns) {
             // DBZ-298 Quoted column names will be sent like that in messages,
             // but stored unquoted in the column names
             final String columnName = Strings.unquoteIdentifierPart(column.getName());
-            undeliveredToastableColumns.remove(columnName);
             int position = getPosition(columnName, table, values);
             if (position != -1) {
-                Object value = column.getValue(() -> (BaseConnection) connection.connection(),
+                Object value = column.getValue(() -> null /* (BaseConnection) connection.connection()*/,
                         connectorConfig.includeUnknownDatatypes());
                 // values[position] = value;
                 values[position] = new Object[]{ value, Boolean.TRUE };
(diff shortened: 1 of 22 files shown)


src/main/java/io/debezium/connector/yugabytedb/connection/YBArray.java
```diff
@@ -0,0 +1,180 @@
+package io.debezium.connector.yugabytedb.connection;
```
vaibhav-yb, line 1: This file is for verifying the changes for the time being. IMO this is not even required as we already send the decoded values from the server, we simply need to parse those values properly which is what the current changes in `YugabyteDBValueConverter` is doing.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeRecordEmitter.java
```diff
@@ -227,17 +222,14 @@ private Object[] updatedColumnValues(List<ReplicationMessage.Column> columns, Ta
...
         for (ReplicationMessage.Column column : columns) {
             // DBZ-298 Quoted column names will be sent like that in messages,
             // but stored unquoted in the column names
             final String columnName = Strings.unquoteIdentifierPart(column.getName());
-            undeliveredToastableColumns.remove(columnName);
 
             int position = getPosition(columnName, table, values);
             if (position != -1) {
-                Object value = column.getValue(() -> (BaseConnection) connection.connection(),
+                Object value = column.getValue(() -> null /*(BaseConnection) connection.connection()*/,
```
vrajat, line 232: Instead of passing `null`, can the parameter be removed. At least from my code browsing, looks like we can edit `ReplicationMessage`

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeRecordEmitter.java
```diff
@@ -284,7 +276,7 @@ private void refreshTableFromDatabase(TableId tableId) {
         try {
             // Using another implementation of refresh() to take into picture the schema information too.
             LOGGER.debug("Refreshing schema for the table {}", tableId);
-            schema.refresh(connection, tableId,
+            schema.refresh(null /* connection */, tableId,
```
vrajat, line 279: Same here. Remove the parameter?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBValueConverter.java
```diff
@@ -916,20 +920,29 @@ protected Object convertArray(Column column, Field fieldDefn, YugabyteDBType ele
             }
             else if (data instanceof PgArray) {
                 try {
-                    final Object[] values = (Object[]) ((PgArray) data).getArray();
+//                  final Object[] values = (Object[]) ((PgArray) data).getArray();
```
vrajat, line 923: nit: Remove the line?

src/main/java/io/debezium/connector/yugabytedb/connection/ReplicationMessageColumnValueResolver.java
```diff
@@ -198,6 +198,6 @@ public static Object resolveValue(String columnName, YugabyteDBType type, String
         }
 
         return value.asDefault(yugabyteDBTypeRegistry, type.getOid(), columnName, fullType,
-                includeUnknownDatatypes, connection);
+                includeUnknownDatatypes, null /* connection - passing null to see if we can makedo without connection */);
```
vrajat, line 201: Remove the parameter if possible?

src/main/java/io/debezium/connector/yugabytedb/connection/pgproto/YbProtoColumnValue.java
```diff
@@ -328,15 +328,15 @@ public Object asArray(String columnName, YugabyteDBType type, String fullType, P
              * }
              */
             final String dataString = asString();
-            return new PgArray(connection.get(), type.getOid(), dataString);
+            return new PgArray(/*connection.get()*/ null, type.getOid(), dataString);
```
vrajat, line 331: Can this return a `YBArray` instead?

src/main/java/io/debezium/connector/yugabytedb/connection/YBArray.java
```diff
@@ -0,0 +1,180 @@
...
+import org.postgresql.util.GT;
+import org.postgresql.util.PSQLException;
+import org.postgresql.util.PSQLState;
+
+import java.sql.SQLException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class YBArray extends PgArray {
```
vrajat, line 18: Are there unit tests that we can also bring in especially for functions like `buildArrayList`

vrajat (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnector.java
```diff
@@ -170,6 +170,16 @@ public List<Map<String, String>> taskConfigs(int maxTasks) {
 
         LOGGER.debug("Configuring {} YugabyteDB connector task(s)", taskConfigs.size());
         closeYBClient();
+
```
vrajat, line 173: Is this better in a `finally` block? One way to avoid a massive whitespace change is to turn this function into a `Impl` function, create the connection, call `Impl` and in `finally` close the connection

src/main/java/io/debezium/connector/yugabytedb/connection/YBArray.java
```diff
@@ -0,0 +1,180 @@
...
+    protected byte @Nullable [] fieldBytes;
+    public YBArray (int oid, String fieldString) throws SQLException {
+        super(null, oid, fieldString);
+        this.oid = oid;
+        this.fieldString = fieldString;
+    }
+
+    protected synchronized YBArrayList buildArrayList(String fieldString) throws SQLException {
+        if (this.arrayList == null) {
+            // TODO Vaibhav: You may need to make the delimiter call dependent on oid
```
suranjan, line 31: This delimeter call we should do in Connector and send to all the tasks.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBValueConverter.java
```diff
@@ -916,20 +920,29 @@ protected Object convertArray(Column column, Field fieldDefn, YugabyteDBType ele
...
     }
 
+    private String[] getArrayElements(Object data) {
+        // This simple hack may not be feasible for 2D arrays
+        // TODO: add test for 2D arrays
+        String dataString = data.toString();
+        dataString = dataString.substring(1, dataString.length() - 1);
+        return dataString.split(",");
+    }
+
```
suranjan, line 945: Instead of this, can't we just resuse the method of pgArray/YBArray by passing the default delimeter.
Ideally it should be able to handle recursion 2d/3d arrays.

Issue #148, opened 2023-02-07
Refactor the changes for stream consistency to smaller PRs


vaibhav-yb:
As per current discussion, the changes are categorized in the following (each bucket requires UTs):
1. Ignoring the safepoint record in the `YugabyteDBStreamingChangeEventSource`
2. Modifications to `add()`, `poll()` and `peek()` functions to properly process tablet safetime
3. Modifications to `Message#compareTo()` i.e. removing the comparison based on sequence
4. Logging related changes

PR #149, opened 2023-02-07
[#148] Ignore safepoint record in default streaming class

vaibhav-yb:


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -379,6 +379,13 @@ protected void getChanges2(ChangeEventSourceContext context,
                                 .getResp()
                                 .getCdcSdkProtoRecordsList()) {
                             CdcService.RowMessage m = record.getRowMessage();
+
```
vrajat, line 382: cc @adithya-yb Is it better if CDC Service in the database sends a SafePoint only if a flag is set?

vaibhav-yb:
This is not required now as there will be server side changes which will instruct the server to not send a SAFEPOINT record in case consistency is not enabled.

PR #150, opened 2023-02-07
[#148] Modifications to add, poll and peek in Merger

vaibhav-yb:
This diff adds the following modifications to the above mentioned functions:
1. `add()` now updates the tablet safetime on adding every message
2. `peek()` now returns a message if the commit time of the front message is **less than or equal to** the stream safetime.
3. Asserts to detect any state which may lead to inconsistency, such asserts would throw an `AssertionError` and will cause the connector to stop now.
4. Unit tests for the added functions.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
@@ -11,6 +11,11 @@ import java.util.*;
  * @author Rajat Venkatesh, Vaibhav Kushwaha
  */
 public class Merger {
+    /*
+        Assumptions made:
+          1. The elements in a merge slot are sorted already because we assume a sorted order
+             from the WAL level itself.
+     */
     private static final Logger LOGGER = LoggerFactory.getLogger(Merger.class);
     private final PriorityQueue<Message> queue = new PriorityQueue<>();
     private final Map<String, List<Message>> mergeSlots = new HashMap<>();
@@ -35,6 +40,18 @@ public class Merger {
         if (message.record.getRowMessage().getOp() == CdcService.RowMessage.Op.SAFEPOINT) {
             LOGGER.debug("Received safe point message {}", message);
             return;
+        }
+
+        // TODO: Wrap these checks under a flag maybe.
+        if (!isMergeSlotSorted(this.mergeSlots.get(message.tablet))) {
+            throw new AssertionError("Merge slot is not sorted");
+        }
+
+        if (!this.mergeSlots.get(message.tablet).isEmpty()
+                && message.commitTime.compareTo(this.mergeSlots.get(message.tablet)
+                .get(this.mergeSlots.get(message.tablet).size() - 1).commitTime) < 0) {
+            throw new AssertionError("Commit time of the newly added message is less than the " +
+                                     "last message in the merge slot");
         }
 
         queue.add(message);
@@ -125,7 +142,7 @@ public class Merger {
         // less than that of the next message in the slot.
         if (!mergeSlots.get(polledMessage.tablet).isEmpty()
             && this.tabletSafeTime.get(polledMessage.tablet)
-                   .compareTo(mergeSlots.get(polledMessage.tablet).get(0).commitTime) == -1) {
+                .compareTo(mergeSlots.get(polledMessage.tablet).get(0).commitTime) < 0) {
                 setTabletSafeTime(polledMessage.tablet, mergeSlots.get(polledMessage.tablet).get(0).commitTime);
         }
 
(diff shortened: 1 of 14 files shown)


src/main/java/io/debezium/connector/yugabytedb/consistent/Message.java
```diff
@@ -49,6 +49,18 @@ public int compareTo(Message o) {
         return 0;
     }
 
+    @Override
+    public boolean equals(Object obj) {
```
vrajat, line 53: Can you add a couple of UTs?

src/main/java/io/debezium/connector/yugabytedb/consistent/Message.java
```diff
@@ -49,6 +49,18 @@ public int compareTo(Message o) {
...
+        Message o = ((Message) obj);
+        return this.tablet.equals(o.tablet)
```
vrajat, line 55: Should tablet and commitTime be sufficient? Also is it common to make sure `this` and `obj` are not the same while comparing two objects in java?

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
```diff
@@ -77,6 +119,16 @@ public synchronized Optional<Message> poll() {
         LOGGER.info("Message is: {}", polledMessage);
         LOGGER.info("Records for tablet: {}", mergeSlots.get(polledMessage.tablet).size());
         mergeSlots.get(polledMessage.tablet).removeIf(item -> item.compareTo(polledMessage) == 0);
+
+        // After removing the record, if there is any record left in the slot then update the tablet
+        // safetime to the value of the first record in the merge slot, iff the tablet safetime is
+        // less than that of the next message in the slot.
+        if (!mergeSlots.get(polledMessage.tablet).isEmpty()
```
vrajat, line 126: IIRC we had decided to remove this block? Is it still required?

vrajat (changes requested):
Will there be a new PR for asserts?

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
vaibhav-yb: Removing this block was still causing the connector getting stuck issue, so this time I added the condition to only update on removal if the first message's commit time is higher than the tablet safetime so that we do not get stuck and that helped.

src/main/java/io/debezium/connector/yugabytedb/consistent/Message.java
vaibhav-yb: Two messages who belong to the same txn for same tablet will have the same tablet ID and commitTime and thus it will not be sufficient.

Not sure about the other question if it's a common practice.

vaibhav-yb:
> Will there be a new PR for asserts?

Yes @vrajat There will be a separate PR for asserts and dumps as it requires me to merge all three of the existing changes first since asserts cover Merger and Messages both.

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
```diff
@@ -19,17 +24,36 @@ public class Merger {
...
+        setTabletSafeTime(message.tablet, message.commitTime);
         if (message.record.getRowMessage().getOp() == CdcService.RowMessage.Op.SAFEPOINT) {
             LOGGER.debug("Received safe point message {}", message);
-            tabletSafeTime.put(message.tablet, message.commitTime);
-            LOGGER.debug("Put {}:{}, verifying {}", message.tablet, message.commitTime, tabletSafeTime.get(message.tablet));
             return;
         }
+
+        // TODO: Wrap these checks under a flag maybe.
+        if (!isMergeSlotSorted(this.mergeSlots.get(message.tablet))) {
```
vrajat, line 46: Is this assert required? Since the next assert checks that the last message is older in terms of sort order, the array should be sorted.

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
```diff
@@ -19,17 +24,36 @@ public class Merger {
...
+            throw new AssertionError("Merge slot is not sorted");
+        }
+
+        if (!this.mergeSlots.get(message.tablet).isEmpty()
```
vrajat, line 50: It is better to compare the two messages and not just the commitTime. I think this condition is not correct as it may break for BEGIN/COMMIT etc. `message.compareTo(lastMessage)` is more appropriate.

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
```diff
@@ -40,6 +64,16 @@ public BigInteger streamSafeTime() {
         return Collections.min(tabletSafeTime.values());
     }
 
+    /**
+     * Set the safetime for the tablet to the given value.
+     * @param tabletId the tablet UUID for which the safetime needs to be set
+     * @param safeTime the safetime to be set
+     */
+    public void setTabletSafeTime(String tabletId, BigInteger safeTime) {
```
vrajat, line 72: I am assuming `setTabletSafeTime` will be called only from `add`. So maybe a duplicate of assert in `add`.

An assert that safeTime can only move forward (or stay in place) will be good to have hear.

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
```diff
@@ -52,9 +86,27 @@ public BigInteger safeTimeForTablet(String tabletId) {
...
     private Optional<Message> peek() {
         Message message = queue.peek();
-        Optional<Message> peeked = message != null && message.commitTime.compareTo(this.streamSafeTime()) < 0
+
+        if (message == null) {
+            LOGGER.warn("Message after peeking is null (actually means to message in queue)");
+        } else {
+            LOGGER.warn("Message is not null in queue - actual message is {}", message);
+            if (!(message.commitTime.compareTo(this.streamSafeTime()) <= 0)) {
+                LOGGER.warn("Comparison commit time for message and stream safetime failed (may indicate issues)");
```
vrajat, line 104: Require a config parameter that after X iterations or time, an assert should be triggered.

src/test/java/io/debezium/connector/yugabytedb/consistent/MessageTest.java
```diff
@@ -61,6 +63,104 @@ public void comparatorTest() {
         assertEquals(1, commit.compareTo(insert));
     }
 
+    @DisplayName("Two unequal messages are detected")
```
vrajat, line 66: This test should also pass the `Parameterized` and `MethodSource` way.

vrajat (changes requested):
For testing asserts, it should be sufficient to add two messages out of order in the same tablet? That should fire the assert (that you may add) in `setTabletSafeTime`.

Two messages that have the same commitTime but out of order like COMMIT followed by BEGIN should trigger the other assert in `add`.

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
```diff
@@ -64,10 +122,17 @@ private Optional<Message> peek() {
         return peeked;
     }
 
+    /**
+     * Get the front message from the queue to process further. This function also removes the
+     * message from the relevant merge slot and update the tablet safetime if required.
```
vrajat, line 127: nit: Can you update the java doc about updating the tablet safetime

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
```diff
@@ -88,4 +162,62 @@ public boolean isEmpty() {
     public boolean isSlotEmpty(String tabletId) {
         return mergeSlots.get(tabletId).isEmpty();
     }
+
+    /**
+     * Whether the messages in the merge slot are sorted properly by the virtue of their
```
vrajat, line 167: Looks like this is unused. Keep it or remove it?

vrajat (commented):
Couple of minor comments. Ack on taking up the config parameter in a different PR

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
vaibhav-yb: Let's keep it for sometime, if required we will remove it around the final cleanup.

src/main/java/io/debezium/connector/yugabytedb/consistent/Merger.java
vaibhav-yb: Done.

vrajat (approved):

PR #154, opened 2023-02-08
[#148] Add UT with complex hierarchical table dependencies

vaibhav-yb:
This PR adds a test with complex hierarchy of tables with FK relationships.

### DIFF
--- a/src/test/java/io/debezium/connector/yugabytedb/YugabyteDBStreamConsistencyTest.java
+++ b/src/test/java/io/debezium/connector/yugabytedb/YugabyteDBStreamConsistencyTest.java
@@ -57,6 +57,8 @@ public class YugabyteDBStreamConsistencyTest extends YugabyteDBTestBase {
     public void before() {
         initializeConnectorTestFramework();
 
+        TestHelper.execute("DROP TABLE IF EXISTS address;");
+        TestHelper.execute("DROP TABLE IF EXISTS contract;");
         TestHelper.execute("DROP TABLE IF EXISTS employee;");
         TestHelper.execute("DROP TABLE IF EXISTS department;");
     }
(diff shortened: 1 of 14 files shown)


src/test/java/io/debezium/connector/yugabytedb/YugabyteDBStreamConsistencyTest.java
```diff
@@ -180,4 +184,130 @@ public void recordsShouldStreamInConsistentOrderOnly() throws Exception {
...
+                            LOGGER.info("Consumed " + totalConsumedRecords.get() + " records");
+                        }
+
+                        return recordsToAssert.size() == total;
+                    });
+        } catch (ConditionTimeoutException exception) {
+            fail("Failed to consume " + totalCount + " records in 600 seconds, consumed " + totalConsumedRecords.get(), exception);
+        }
+
+        assertEquals(total, recordsToAssert.size());
```
suranjan, line 293: Do we just assert the count?

suranjan (changes requested):

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBStreamConsistencyTest.java
vaibhav-yb: No, we assert the record's position as well later in the test.

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBStreamConsistencyTest.java
```diff
@@ -180,4 +184,130 @@ public void recordsShouldStreamInConsistentOrderOnly() throws Exception {
...
+        // Dummy wait
+        TestHelper.waitFor(Duration.ofSeconds(25));
+
+        List<SourceRecord> recordsToAssert = new ArrayList<>();
+
+        final long total = totalCount;
+        AtomicLong totalConsumedRecords = new AtomicLong();
+        try {
+            Awaitility.await()
+                    .atMost(Duration.ofSeconds(600))
```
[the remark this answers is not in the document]
vaibhav-yb: This is what we refactored our iterative functions to, if it remains a function of iteration then it will forever keep waiting for records if it gets stuck. We could have added another timeout in that iteration but then using Awaitility seemed a better idea - all the other tests currently follow the same.

PR #155, opened 2023-02-15
Adding changes for consistent streaming of records

vaibhav-yb:
This PR adds the support for consistent streaming of the change records to the connector.


src/main/java/io/debezium/connector/yugabytedb/SourceInfo.java
```diff
@@ -110,24 +119,25 @@ protected SourceInfo(YugabyteDBConnectorConfig connectorConfig, OpId lastCommitL
...
      * @return this instance
      */
-    protected SourceInfo update(String tabletId, OpId lsn, Instant commitTime, String txId,
-                                TableId tableId,
-                                Long xmin) {
+    protected SourceInfo update(String tabletId, OpId lsn, long commitTime, String txId,
+                                TableId tableId, Long xmin, Long recordTime) {
         this.lsn = lsn;
-        if (commitTime != null) {
-            this.timestamp = commitTime;
```
vrajat, line 127: `timestamp` is not set anymore. Is it being used anywhere?

vrajat (commented):
`SourceInfo` and `YugabyteDBOffsetContext` and their usage requires unit tests in IMO. 

Similarly I think `YugabyteDBStreamingChangeEventSource` require some basic unit tests. Most of the changes are moving from private to protected. However, as more changes are made, it should be possible to add tests quickly.

src/main/java/io/debezium/connector/yugabytedb/SourceInfo.java
vaibhav-yb: It's not being used anywhere, but it gets published as a part of the message in the topic. I can we can still populate the field.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
```diff
@@ -214,14 +214,14 @@ public void updateSnapshotPosition(Instant timestamp, TableId tableId) {
     }
 
     public void updateWalPosition(String tabletId, OpId lsn, OpId lastCompletelyProcessedLsn,
-                                  Instant commitTime,
-                                  String txId, TableId tableId, Long xmin) {
+                                  long commitTime,
+                                  String txId, TableId tableId, Long xmin, Long recordTime) {
```
vaibhav-yb, line 218: @vrajat Should this be `BigInteger` instead of `Long` or `long`

vrajat (commented):
Changes look good in general. Can you provide a description of why the refactor is useful? Is it removing the duplicate or unused code or cleaning up the APIs?

vrajat:
Sorry - ignore the previous comment. Wrong diff.

vrajat:
What is the plan to merge this PR? A couple of action items are to ensure github actions and jenkins runs are successful

vaibhav-yb:
@vrajat I have modified the GH Actions accordingly in order to get them run successfully, but the only issue will be that the consistency tests will fail in the runs against the stable YB build as that runs against 2.16 and doesn't have the required changes. I will update that to not run the test for consistency so that even they are successful.

Are there any other items you're talking about?

vrajat:
Will jenkins run the consistency tests?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
```diff
@@ -0,0 +1,457 @@
...
+                }
+                catch (InterruptedException ie) {
+                    LOGGER.warn("Connector retry sleep interrupted by exception: {}", ie);
+                    Thread.currentThread().interrupt();
+                }
+            }
+        }
+    }
+
+    private void dispatchMessage(YugabyteDBOffsetContext offsetContext, Map<String, Boolean> schemaNeeded,
```
vrajat, line 308: This function should be reusable from Streaming Source as well? Can you add an issue to track the task of reusing this fn?

vrajat (approved):

PR #160, opened 2023-02-15
[#148] Ignore DDL records while adding to merger

vaibhav-yb:
This PR adds the diff which will directly get the DDL records dispatched and not let them get added to the merger.


src/main/java/io/debezium/connector/yugabytedb/YugabyteDbConsistentStreaming.java
```diff
@@ -202,11 +202,18 @@ protected void getChanges2(ChangeEventSourceContext context,
                                     .getCdcSdkProtoRecordsList()) {
                                 CdcService.RowMessage.Op op = record.getRowMessage().getOp();
 
-                                merger.addMessage(new Message.Builder()
-                                        .setRecord(record)
-                                        .setTabletId(tabletId)
-                                        .setSnapshotTime(response.getSnapshotTime())
-                                        .build());
+                                if (!schemaNeeded.get(tabletId) && record.getRowMessage().getOp() == CdcService.RowMessage.Op.DDL) {
```
vrajat, line 205: Why is `!schemaNeeded.get(tabletId)`? Should all DDL bypass merger?

vrajat (approved):

PR #161, opened 2023-02-20
[#162] Support for EXPLICIT checkpointing

suranjan:
Code changes to implement EXPLICIT checkpointing on the connector so that when Kafka Connect invokes the `commitOffset()` callback, we can use the same offsets to commit the checkpoint on the server.

For EXPLICIT checkpointing, the following code flow is used:
1. Kafka Connect invokes the callback `commit()` which further invokes `commitOffset(map-of-offsets)` in the Debezium API
2. Both the streaming and snapshot change event source classes maintain a map `tabletToExplicitCheckpoint` which stores the offsets sent by Kafka Connect.
3. So when the connector gets the acknowledgement back by saying that Kafka has received records till certain offset, we update the value in `tabletToExplicitCheckpoint`
4. While making the next `GetChanges` call, we pass the value from the map `tabletToExplicitCheckpoint` for the relevant tablet and hence the server then takes care of updating those checkpointed values in the `cdc_state` table.


src/main/java/io/debezium/connector/yugabytedb/YugabyteDbConsistentStreaming.java
```diff
@@ -279,6 +279,11 @@ beginCountForTablet, message.tablet, new YBPartition(message.tablet),
         }
     }
 
+    @Override
+    public void commitOffset(Map<String, ?> offset) {
```
vrajat, line 283: Is the method required as this function does not have different behaviour to the base class?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -659,25 +660,50 @@ protected void maybeWarnAboutGrowingWalBacklog(boolean dispatched) {
...
-        ReplicationStream replicationStream = null; // this.replicationStream.get();
-        final OpId commitLsn = null; // OpId.valueOf((String) offset.get(PostgresOffsetContext.LAST_COMMIT_LSN_KEY));
-        final OpId changeLsn = OpId.valueOf((String) offset.get(YugabyteDBOffsetContext.LAST_COMPLETELY_PROCESSED_LSN_KEY));
-        final OpId lsn = (commitLsn != null) ? commitLsn : changeLsn;
-
-        if (replicationStream != null && lsn != null) {
-            if (LOGGER.isDebugEnabled()) {
-                LOGGER.debug("Flushing LSN to server: {}", lsn);
+        try {
+            LOGGER.info("Committing offsets on server");
```
vrajat, line 664: Will `StreamingChangeEventSource` also use explicit checkpoints? I assumed implicit checkpoints are sufficient in the default mode.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDbConsistentStreaming.java
vaibhav-yb: You're right, I have removed the method.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
vaibhav-yb: It can, that was already on our roadmap of implementing things for the connector. One thing we can do it add a check/flag in the connector to only call this method when the stream is created in the `EXPLICIT` mode on the server side, otherwise we will not be committing any offsets.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
```diff
@@ -32,7 +34,12 @@
 
 public class YugabyteDBOffsetContext implements OffsetContext {
     public static final String LAST_COMPLETELY_PROCESSED_LSN_KEY = "lsn_proc";
-    public static final String LAST_COMMIT_LSN_KEY = "lsn_commit";
+
+    // TODO: Remove commented area before final merge (commenting because this is not being used)
```
vaibhav-yb, line 38: Note to self: Remove this line.

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -247,6 +248,47 @@ public static OpId getOpIdFromGetTabletListResponse(GetTabletListToPollForCDCRes
     return null;
   }
 
+    /**
```
vrajat, line 251: The design of this functions seems convoluted. The ultimate goal is to get CDCStreamInfo. However to get that, a table UUID is required. So two rpc calls are required. Not a perf concern but a convoluted logic concern. Why doest `getDBStreamInfoResponse` return `CDCStreamInfo` as well?

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -247,6 +248,47 @@ public static OpId getOpIdFromGetTabletListResponse(GetTabletListToPollForCDCRes
...
+     * Get the {@link CDCStreamInfo} object for the specified configuration. Note that the connector
+     * configuration is just used to extract the stream ID
```
vrajat, line 253: Nit: ConnectorConfig is also used to instantiate the YBClient?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
```diff
@@ -146,34 +145,20 @@ public static YugabyteDBOffsetContext initialContext(YugabyteDBConnectorConfig c
...
-            result.put(SourceInfo.SNAPSHOT_KEY, true);
-            result.put(SourceInfo.LAST_SNAPSHOT_RECORD_KEY, lastSnapshotRecord);
-        }
-        if (lastCompletelyProcessedLsn != null) {
-            result.put(LAST_COMPLETELY_PROCESSED_LSN_KEY, lastCompletelyProcessedLsn.toSerString());
-        }
-        if (lastCommitLsn != null) {
-            result.put(LAST_COMMIT_LSN_KEY, lastCommitLsn.toSerString());
+
+        for (Map.Entry<String, SourceInfo> entry : this.tabletSourceInfo.entrySet()) {
```
vrajat, line 149: Is this storing lsn for every tablet instead of fields in SourceInfo?

vaibhav-yb: Yes.

vrajat (approved):
Please add a description in the commit and in the code to explain the control flow. A few bullet points are sufficient.

PR #175, opened 2023-03-07
Do not run consistency related tests against stable YB build

vaibhav-yb:
This PR adds the consistency related tests to the list of tests **NOT** to be run against the stable builds (currently 2.16) of YugabyteDB since the server side changes required for them are not present in the relevant 2.16 build being used for testing.

### DIFF
--- a/.github/workflows/sanity-workflow.yml
+++ b/.github/workflows/sanity-workflow.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Sanity Test that code compiles and tests pass.
         env:
           YB_DOCKER_IMAGE: ${{ secrets.YB_DOCKER_IMAGE }}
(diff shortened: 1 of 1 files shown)


vrajat:
@vaibhav-yb should we consider running the github actions in a runner? There is a runner available. It is possible to download and compile the right version of yugabytedb. Also run against all stable versions.

vaibhav-yb:
We can consider that path as well. Currently I have to go and manually update the GitHub secret to point to the latest & correct version for both preview and stable builds.

vrajat (commented):
Minor style suggestion. Is using maven profiles better?

vaibhav-yb:
Closing this since we now have annotation support where we can specify whether to run the build with preview or stable branch.

PR #178, opened 2023-03-07
Test addition and code modification for EXPLICIT checkpointing

vaibhav-yb:
This PR introduces the following changes to the codebase:
1. Addition of a UT to verify the explicit checkpointing behaviour of the connector
2. The `YugabyteDBChangeEventSourceCoordinator` will only call `commitOffsets()` for snapshot iff snapshot is enabled, it will ignore `commitOffsets` for snapshot in case snapshot is not enabled.
3. Fixes a bug in verifying the value for the option of EXPLICIT checkpoint under `YBClientUtils`


vrajat (approved):

vaibhav-yb:
Abandoning this PR as the changes have landed as a part of another PR #179

PR #182, opened 2023-03-14
[#181] Log test names with each test

### LINKED ISSUE #181: Add log to print test names to each test
Today, if we run a test class, then all the tests are run and at the end the list of failed tests is printed. However, let's say something failed in between, then it is not possible to deduce which test failed just by looking at the logs.

To facilitate that, add a log stating that we are starting this particular test now. eg.

```java
public void testSomethingImportant() {
  LOGGER.info("Starting testSomethingImportant");
}
```

vaibhav-yb:
Log test names at the start of every test in order to make debugging easier in case of failures by figuring out the exact test name from the log files.

Closes #181

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
         <protobuf.output.directory>${project.basedir}/generated-sources</protobuf.output.directory>
 
         <!-- Test -->
-        <version.junit>5.9.0</version.junit>
+        <version.junit>5.9.2</version.junit>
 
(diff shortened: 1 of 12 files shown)


vrajat (changes requested):
You should instead implement a TestExecutionListener: https://junit.org/junit5/docs/current/user-guide/#running-tests-listeners

vrajat (approved):
Nice! Much cleaner

PR #186, opened 2023-03-21
[#183] Throw fatal exceptions instead of simply catching

### LINKED ISSUE #183: Throw exceptions rather than simply logging and moving ahead
There are places in the code where we simply log the error and move ahead without even throwing it out properly. We want those errors to be thrown so that the connector startup fails early.

vaibhav-yb:
There was one fatal exception at the top connector level which was simply logged and not thrown resulting in the connector moving ahead with the processing. The exception will not be properly thrown as well.

Additionally, this closes #183

### DIFF
--- a/src/test/java/io/debezium/connector/yugabytedb/ClientAPITest.java
+++ b/src/test/java/io/debezium/connector/yugabytedb/ClientAPITest.java
@@ -4,7 +4,6 @@ import io.debezium.connector.yugabytedb.common.YugabyteDBContainerTestBase;
 import org.junit.jupiter.api.*;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
-import org.yb.cdc.CdcService;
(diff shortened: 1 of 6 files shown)


src/test/java/io/debezium/connector/yugabytedb/ClientAPITest.java
```diff
@@ -0,0 +1,81 @@
...
+
+		try {
+			// Now get the tablet list for all the tables.
+			for (YBTable table : tables) {
+				assertNotNull(table);
+				GetTabletListToPollForCDCResponse resp =
+						ybClient.getTabletListToPollForCdc(table, dbStreamId, table.getTableId());
+				assertNotNull(resp);
+			}
+		} catch (Exception e) {
```
vrajat, line 76: This block is not required. The Junit will fail it if there is an exception.

vrajat (approved):
Approving it. However you should remove the try/catch in the block.

vaibhav-yb:
Merging after approving the review comments.

PR #187, opened 2023-03-21
Add new source for processing snapshot of colocated tables

vaibhav-yb:
This PR adds a custom source class which would help in processing the snapshot for colocated tables. Major changes are:
1. OpIds for colocated tables are keyed by `tableId.tabletId` instead of just `tabletId`
2. All the colocated tables for the same tablet will be grouped into one task. This also adds a requirement that if there are multiple colocated tablets the `task.max` property should be explicitly provided as equal to the number of unique tablets. In case there is a non-colocated tablet as well, then the property should be at least 1 greater than the number of unique tablets.
3. Added an internal config to indicate whether a task has only colocated tables.


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBColocatedSnapshotChangeEventSource.java
```diff
@@ -0,0 +1,579 @@
...
+        return SnapshotResult.aborted();
+    }
+
+    /**
+     * This function returns a lookup key to get the values in different map based structures.
+     * @param tableId the table UUID to use while forming lookup key
+     * @return a string value for the lookup key of the format {@code tableId.colocatedTabletId}
+     * @throws NullPointerException if {@code colocatedTabletId} has not been populated yet
+     */
+    private String getLookupKey(String tableId) throws NullPointerException {
```
vaibhav-yb, line 419: Note: Throw IllegealStateException

vrajat:
Is this PR still valid? It has conflicts with the main branch

PR #191, opened 2023-03-28
[#194] Store OpIds and offset contexts against a lookup key

vaibhav-yb:
This PR changes the mechanism with which we store the OpIds, OffsetContext, SourceInfo and all the related things. Before this, they were keyed against the `tabletId` but this change modifies the logic so that they can be queried with `tableId.tabletId`. This will also help us in accommodating the changes for streaming the snapshot for colocated tables.

For example:
**Older logic:***
```
tablet_1 : op_id_1,
tablet_2 : op_id_2
```
**New logic:**
```
table.tablet_1 : op_id_1,
table.tablet_2 : op_id_2
```

The major changes in this PR (in no specific order) are:
1. Lookup of data using `tableId.tabletId` key
2. Sending a `snapshot_done_key` just once in the `GetChanges` call after we receive the snapshot complete marker from the server
3. Modification of existing snapshot tests to make them parameterised so that they now run twice - once with colocated tables, once without colocated tables
4. Addition of a test to ensure the snapshot with the mix of colocated and non colocated tables
5. Addition of a grouping function `YugabyteDBConnectorUtils#groupPartitionsSmartly` (and tests) which will ensure that the tables belonging to the same colocated tablet always go to the same task.

However, there is **one exception** to the above:
In streaming mode, we are now forcefully marking all the tables as non-colocated and in snapshot we are marking them all as colocated.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
@@ -519,8 +519,8 @@ public class YugabyteDBStreamingChangeEventSource implements
                                             && dispatcher.dispatchDataChangeEvent(part, tableId, new YugabyteDBChangeRecordEmitter(part, offsetContext, clock, connectorConfig,
                                                     schema, connection, tableId, message, pgSchemaNameInRecord, tabletId, taskContext.isBeforeImageEnabled()));
 
-                                    if (recordsInTransactionalBlock.containsKey(entry.getKey() + "." + tabletId)) {
-                                        recordsInTransactionalBlock.merge(entry.getKey() + "." + tabletId, 1, Integer::sum);
+                                    if (recordsInTransactionalBlock.containsKey(part.getId())) {
+                                        recordsInTransactionalBlock.merge(part.getId(), 1, Integer::sum);
                                     }
 
                                     maybeWarnAboutGrowingWalBacklog(dispatched);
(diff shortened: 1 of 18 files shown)


suranjan:
> Sending a snapshot_done_key just once in the GetChanges call after we receive the snapshot complete marker from the server
Do we handle this for explicit and only when we get the callback from kafka/commitOffset we set snapshot complete?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -511,15 +512,15 @@ else if (message.isDDLMessage()) {
...
+                                    offsetContext.updateWalPosition(part, lsn, lastCompletelyProcessedLsn, message.getCommitTime(),
                                             String.valueOf(message.getTransactionId()), tableId, null/* taskContext.getSlotXmin(connection) */);
 
                                     boolean dispatched = message.getOperation() != Operation.NOOP
                                             && dispatcher.dispatchDataChangeEvent(part, tableId, new YugabyteDBChangeRecordEmitter(part, offsetContext, clock, connectorConfig,
                                                     schema, connection, tableId, message, pgSchemaNameInRecord, tabletId, taskContext.isBeforeImageEnabled()));
 
-                                    if (recordsInTransactionalBlock.containsKey(tabletId)) {
-                                        recordsInTransactionalBlock.merge(tabletId, 1, Integer::sum);
+                                    if (recordsInTransactionalBlock.containsKey(entry.getKey() + "." + tabletId)) {
```
vrajat, line 522: Should `tabletId` still be used as key as streaming code does not care about table ids?

vrajat (approved):

PR #195, opened 2023-03-29
[#192] Send a snapshot done key at the beginning of every streaming iteration

### LINKED ISSUE #192: Introduce a snapshot end marker to verify the one from server side
We need to add a `SNAPSHOT_DONE_KEY` which would help us to verify whether the OpId we have received marks the end of the snapshot.

vaibhav-yb:
This PR modifies the code logic so that at the beginning of every iteration for streaming, we send out a key which has the `snapshot_done_key` as its key to tell the server that it is the point where snapshot has been completed.

Additionally, this closes #192

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
         <version.kafka>3.3.1</version.kafka>
         <version.org.slf4j>1.7.36</version.org.slf4j>
         <version.logback>1.4.0</version.logback>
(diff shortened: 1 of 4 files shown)


vaibhav-yb:
Not needed, these changes were required for colocated table snapshot and will go in that PR only.

PR #201, opened 2023-03-30
[#200] Add a transformer for Striim compatibility

### LINKED ISSUE #200: Create a transformer for striim
The objective of this issue is to create a transformer so that the records sent to kafka are compatible with [striim](https://www.striim.com/docs/en/postgresql-cdc.html#UUID-bb4c7685-acc1-6f6a-f4a2-afbfecf4febf).

nishantwrp:
Create a transformer so that the records sent to kafka are compatible with [striim](https://www.striim.com/docs/en/postgresql-cdc.html#UUID-bb4c7685-acc1-6f6a-f4a2-afbfecf4febf).

### Target Structure

**metadata (struct)** 
- LSN: log sequence number of the transaction's commit
- NEXT_LSN: next log sequence number (used for reconnecting to the replication slot after a non-fatal network interruption)
- OperationName: INSERT, UPDATE, or DELETE
- PK_UPDATE: included only when an UPDATE changes the primary key
- Sequence: incremented for each operation within a transaction
- TableName: the name of the table including its schema
- TimeStamp: timestamp from the replication subscription
- TxnID: transaction identifier

**data (array)**
- for an INSERT operation, the values that were inserted
- for an UPDATE, the values after the operation was completed
- for a DELETE, the value of the primary key and nulls for the other fields

**before (array)**
- for UPDATE operations, contains the primary key value from before the update. 

Closes #200.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
@@ -1,0 +1,208 @@
+package io.debezium.connector.yugabytedb.transforms;
+
+import java.util.*;
+
+import io.debezium.connector.yugabytedb.transforms.SchemaUtil;
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.connect.connector.ConnectRecord;
+import org.apache.kafka.connect.data.Field;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.Schema.Type;
+import org.apache.kafka.connect.data.SchemaBuilder;
+import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.connect.transforms.Transformation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.yb.util.Pair;
+
+public class StriimCompatible<R extends ConnectRecord<R>> implements Transformation<R> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(StriimCompatible.class);
+
+    @Override
+    public R apply(final R record) {
+        if (record == null || (record.value() != null && !(record.value() instanceof Struct))) {
+            return record;
+        }
+
+        List<String> primaryKeys = getAllFieldsInOrder(record.keySchema());
+
+        Schema updatedSchemaForValue = null;
+        Struct updatedValueForValue = null;
+        if (record.value() != null) {
+            Pair<Schema, Struct> val = getUpdatedValueAndSchema(record.valueSchema(), (Struct) record.value(), primaryKeys);
+            updatedSchemaForValue = val.getFirst();
+            updatedValueForValue = val.getSecond();
+        }
+
+        return record.newRecord(record.topic(), record.kafkaPartition(), record.keySchema(), record.key(), updatedSchemaForValue, updatedValueForValue, record.timestamp());
+    }
+
+    @Override
+    public ConfigDef config() {
+        return new ConfigDef();
+    }
+
+    @Override
+    public void close() {
+    }
+
+    private List<String> getAllFieldsInOrder(Schema schema) {
+        List<String> fields = Arrays.asList(new String[schema.fields().size()]);
+        for (Field field : schema.fields()) {
+            fields.set(field.index(), field.name());
+        }
+        return fields;
+    }
+
+    private boolean isValueSetStruct(Field field) {
+        return field.schema().fields().size() == 2
+                && (Objects.equals(field.schema().fields().get(0).name(), "value")
+                && Objects.equals(field.schema().fields().get(1).name(), "set"));
+    }
+
+    private Schema makeMetadataSchema() {
+        final SchemaBuilder builder = SchemaBuilder.struct();
+        builder.field("LSN", Schema.STRING_SCHEMA);
+        builder.field("OperationName", Schema.STRING_SCHEMA);
+        builder.field("PK_UPDATE", Schema.OPTIONAL_BOOLEAN_SCHEMA);
+        builder.field("Sequence", Schema.STRING_SCHEMA);
+        builder.field("TableName", Schema.STRING_SCHEMA);
+        builder.field("TxnID", Schema.STRING_SCHEMA);
+        return builder.build();
+    }
+
+    private Struct makeMetadata(Struct value, Schema metadataSchema) {
+        Struct metadata = new Struct(metadataSchema);
+        Struct sourceValue = (Struct) value.get("source");
+        metadata.put("LSN", sourceValue.get("lsn").toString());
+        metadata.put("Sequence", sourceValue.get("sequence").toString());
+        metadata.put("TxnID", sourceValue.get("txId"));
+        metadata.put("TableName", sourceValue.get("schema") + "." + sourceValue.get("table"));
+        return metadata;
+    }
+
+    private Schema makeUpdatedSchema(Schema schema) {
+        final SchemaBuilder arrayBuilder = SchemaBuilder.array(Schema.OPTIONAL_STRING_SCHEMA);
+        final SchemaBuilder builder = SchemaUtil.copySchemaBasics(schema, SchemaBuilder.struct());
+        builder.field("metadata", makeMetadataSchema());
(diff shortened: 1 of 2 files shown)


src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
```diff
@@ -0,0 +1,208 @@
...
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.Schema.Type;
+import org.apache.kafka.connect.data.SchemaBuilder;
+import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.connect.transforms.Transformation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.yb.util.Pair;
+
+public class StriimCompatible<R extends ConnectRecord<R>> implements Transformation<R> {
```
vaibhav-yb, line 18: Add a java doc comment explaining the use of this transformer in brief. For reference, see other classes in the repo.

src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
```diff
@@ -0,0 +1,208 @@
...
+            return values;
+        }
+
+        for (Field field : value.schema().fields()) {
+            LOGGER.debug("Considering value {}", field.name());
+            if (field.schema().type() == Type.STRUCT) {
+                LOGGER.debug("Value is a struct");
+                Struct fieldValue = (Struct) value.get(field);
+                if (isValueSetStruct(field)) {
+                    LOGGER.debug("value is valueset");
```
vaibhav-yb, line 106: The log doesn't make much sense if seen from the bigger picture, can you please improve it a little? Maybe mention that `Value while extracting data is a valueset` - or something similar.

src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
```diff
@@ -0,0 +1,208 @@
...
+        LOGGER.debug("Updated schema as json: " + io.debezium.data.SchemaUtil.asString(updatedSchema));
+
+        List<String> allFields = getAllFieldsInOrder(schema.field("after").schema());
+
+        LOGGER.debug("Original value as json: {}", io.debezium.data.SchemaUtil.asDetailedString(value));
+        Struct newVal = new Struct(updatedSchema);
+        Struct metadata = makeMetadata(value, updatedSchema.field("metadata").schema());
+        newVal.put("columns", allFields);
+
+        switch ((String)value.get("op")) {
```
vaibhav-yb, line 160: `(String) value.get("op")` --> `value.getString("op")`

src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
```diff
@@ -0,0 +1,208 @@
...
+                removeNonPrimaryKeyValues(oldValues, primaryKeys);
+
+                metadata.put("OperationName", "DELETE");
+                newVal.put("metadata", metadata);
+                newVal.put("data", convertToOrderedList(oldValues, allFields));
+
+                LOGGER.debug("Update value as json: {}", io.debezium.data.SchemaUtil.asDetailedString(newVal));
+                return new org.yb.util.Pair<>(updatedSchema, newVal);
+            }
+            default: {
```
vaibhav-yb, line 198: @suranjan Are we going to need a case for `READ (r)` as well?

src/test/java/io/debezium/connector/yugabytedb/transforms/StriimCompatibleTest.java
```diff
@@ -0,0 +1,245 @@
...
+import org.apache.kafka.connect.source.SourceRecord;
+import org.junit.jupiter.api.Test;
+
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+
+public class StriimCompatibleTest {
```
vaibhav-yb, line 16: Add a javadoc for this test as well. Take reference from other classes.

vaibhav-yb (changes requested):

src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
suranjan: Yes, we should convert snapshot entry as well.

src/test/java/io/debezium/connector/yugabytedb/transforms/StriimCompatibleTest.java
nishantwrp: Done.

src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
nishantwrp: done

src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
```diff
@@ -0,0 +1,208 @@
...
+        builder.field("Sequence", Schema.STRING_SCHEMA);
+        builder.field("TableName", Schema.STRING_SCHEMA);
+        builder.field("TxnID", Schema.STRING_SCHEMA);
+        return builder.build();
+    }
+
+    private Struct makeMetadata(Struct value, Schema metadataSchema) {
+        Struct metadata = new Struct(metadataSchema);
+        Struct sourceValue = (Struct) value.get("source");
+        metadata.put("LSN", sourceValue.get("lsn").toString());
```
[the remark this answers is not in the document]
nishantwrp: done

src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
nishantwrp: done

src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
nishantwrp: Removed these logs because there were logs already for printing the complete updated value. Actually, I added these logs originally when I was testing this transformer.

src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
nishantwrp: Actually, there is no documentation on read records [here](https://www.striim.com/docs/en/postgresql-cdc.html#UUID-bb4c7685-acc1-6f6a-f4a2-afbfecf4febf). So, I am unsure about what to send in the `data`, `metadata` field.

src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
```diff
@@ -0,0 +1,216 @@
...
+public class StriimCompatible<R extends ConnectRecord<R>> implements Transformation<R> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(StriimCompatible.class);
+
+    @Override
+    public R apply(final R record) {
+        if (record == null || (record.value() != null && !(record.value() instanceof Struct))) {
+            return record;
+        }
+
+        List<String> primaryKeys = getAllFieldsInOrder(record.keySchema());
```
vrajat, line 32: Nit: Is the variable name correct? Should it be called `columnNames` instead?

vrajat (approved):

src/main/java/io/debezium/connector/yugabytedb/transforms/StriimCompatible.java
nishantwrp: Oh, I thought the `columnNames` in the key will be primary keys only. Isn't that the case @vrajat?

vaibhav-yb: I believe that is the case, the columns in keySchema are primary keys.

vaibhav-yb (approved):

PR #202, opened 2023-03-31
[CDCSDK#16556] Check for snapshot completion after GetChanges call

vaibhav-yb:
The issue arose where we were checking whether the `OpId` is the snapshot complete marker on the `from_op_id` i.e. `cp` in our case - that is prone to errors as reported in stress tests as we may end up calling `GetChanges` with a checkpoint meant for streaming.

One such error occured in the call as mentioned in the ticket yugabyte/yugabyte-db#16556, on investigation, it was found that the flow was like:
1. Call `GetChanges` on `from_op_id` (for snapshot)
2. Receive a end marker in response and update the LSN
3. Since we were checking for snapshot end marker on `from_op_id` - it got ignored
4. The LSN now would not be the one we need to use for, but instead it would be the one being used for streaming
5. Now if we call `GetChanges` and if the tablet got split, we will throw a "tablet split detected" error - which will not be handled at the snapshot level - thus leading to the failure of the task.

This PR fixes the above error by checking the marker on the response OpId instead of request OpId and there are following cases where we check:
| **EXPLICIT** | **IMPLICIT** |
| :--- | :--- |
| If the explicit checkpoint is snapshot end marker, add tablet to snapshot completion list | If checkpoint received in response is the snapshot end marker then add tablet to snapshot completion list |
| If explicit checkpoint is not end marker and we have not received the callback yet, add the tablet to a waitlist so that we can avoid calling `GetChanges` for it in further iterations | |

For reference, the snapshot end marker is an `OpId` with the following values:
`term` : some term value
`index` : some index value
`key` : empty byte array
`write_id` : 0
`time` : 0

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -372,18 +372,6 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
                     connectorConfig.streamId(), tabletId, cp.getTerm(), cp.getIndex(), cp.getKey(), 
                     cp.getWrite_id(), cp.getTime(), schemaNeeded.get(tabletId),
                     taskContext.shouldEnableExplicitCheckpointing() ? tabletToExplicitCheckpoint.get(tabletId) : null);
-
-                // If EXPLICIT checkpointing is enabled then check if the checkpoint is the marker for snapshot completion
-                // and in case it is IMPLICIT checkpointing, the marker value should be checked on the from_op_id we are sending
-                // to the server.
-                if ((taskContext.shouldEnableExplicitCheckpointing()
-                        && isSnapshotCompleteMarker(OpId.from(tabletToExplicitCheckpoint.get(tabletId))))
-                            || isSnapshotCompleteMarker(cp)) {
-                  // This will mark the snapshot completed for the tablet
-                  snapshotCompletedTablets.add(tabletId);
-                  LOGGER.info("Snapshot completed for tablet {} belonging to table {} ({})",
-                          tabletId, table.getName(), tableId);
-                }
                 
                 // Process the response
                 for (CdcService.CDCSDKProtoRecordPB record : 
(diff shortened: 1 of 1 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -336,12 +336,17 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
...
+                // that we do not end up sending redundant GetChanges calls.
+                Set<String> tabletsWaitingForCallback = new HashSet<>();
+
                 String tableId = tableIdToTabletId.getKey();
                 YBTable table = tableIdToTable.get(tableId);
 
                 String tabletId = tableIdToTabletId.getValue();
                 YBPartition part = new YBPartition(tabletId);
-                
+
```
vaibhav-yb, line 349: These blank lines are the result of IntelliJ auto-formatter and it's removing extra blank spaces automatically, somehow I am unable to get them removed, please ignore them.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -460,10 +458,48 @@ && isSnapshotCompleteMarker(OpId.from(tabletToExplicitCheckpoint.get(tabletId)))
...
+                      b. If the explicit checkpoint is not the snapshot end marker, there is
+                         a possibility that the finalOpId received above in the response is the
+                         snapshot complete marker - add the tablet to the set so that we do not end
+                         up calling redundant GetChanges
+                   2. Implicit checkpointing:
+                      a. In this case, only checking the response final_op_id is enough to mark
+                         the snapshot as completed.
+                 */
+                if (taskContext.shouldEnableExplicitCheckpointing()) {
+                  if (isSnapshotCompleteMarker(OpId.from(this.tabletToExplicitCheckpoint.get(tabletId)))) {
```
vrajat, line 484: Why not always add the tabletId to `snapshotCompletedTablets`? In that case, the callback is a no-op which should not matter.

vrajat (commented):
Very nice description. Is it possible to add tests here or in `cdcsdk-testing` or is it too hard to reproduce the issue?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
vaibhav-yb: That is because `snapshotCompletedTablets` contain the list of the tablets for which the following two conditions are met:
1. The server has send the snapshot end marker.
2. In case of EXPLICIT checkpointing - Kafka has sent the callback so we are sure we have received the data.

The additional set is for cases of EXPLICIT checkpointing only where point 2 is not satisfied, so we know that server has send the data (1 is satisfied) but Kafka hasn't acknowledged the message's presense. If we always add the tabletId to `snapshotCompletedTablets` - there is a chance that when the connector crashes, we may lose some data since we may not have published them to Kafka yet.

vaibhav-yb: Re: adding tests - it's a little difficult to reproduce the issue in our tests, I spent some time but perhaps not enough to keep manipulating the tests. It was getting reproduced in the stress runs easily, which is now passing too.

Reference run: http://stress.dev.yugabyte.com/stress_test/c305a9d5-1096-45b3-a3ef-0bb39aa9ec5f

vrajat: Can you add this explanation as comments in the  EXPLICIT section?

vrajat (approved):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
vaibhav-yb: Added.

PR #210, opened 2023-04-19
[yugabyte/yugabyte-db#16905] Fix metrics regex 

vaibhav-yb:
This PR modifies the regex to use the partition id instead of just tablet id which we further use to scrape Debezium metrics for Prometheus.

This closes yugabyte/yugabyte-db#16905

### DIFF
--- a/metrics.yml
+++ b/metrics.yml
@@ -113,14 +113,14 @@ rules:
       context: "$2"
       table: "$4"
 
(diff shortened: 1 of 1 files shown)


vrajat:
Have you manually tested it?

vaibhav-yb:
Yes, I pushed a latest change which changed the label from `id` to `partition` - now we are good. I have tested it manually and verified the metrics.

We will require some changes in `yugabyte/cdc-examples` but that would be a separate PR there.

PR #212, opened 2023-04-24
[yugabyte/yugabyte-db#16929] Add method to implement yugabyted restart

vaibhav-yb:
This PR adds the method to restart the `yugabyted` process inside the YugabyteDB container. There are 3 methods added to facilitate the process related operations:
1. `stopYugabyteDB()` - stops the yugabyted process
2. `startYugabyteDB()` - starts the yugabyted process
3. `restartYugabyteDB(long)` - restarts the yugabyted process with a wait of specified milliseconds between stopping and starting the process.

A test has also been added to ensure that the APIs are functioning as expected, to run the test, use the command:
```
mvn clean test -Dtest=YugabyteDBDatatypesTest#verifyConsumptionAfterRestart
```

To achieve this, the following changes have been made:
1. Container now starts with a command for infinite loop - this ensured that even if we stop the `yugabyted` process, we do not end up exiting from the container.
2. Overridden a super function from the `YugabyteYSQLContainer` in a new class i.e. `YugabyteCustomContainer` to skip the execution of `waitUntilContainerIsReady` method which was waiting for a successful JDBC connection, but since we are not starting the yugabyted process in the beginning, we cannot open a connection and will need to skip this method
    i. We can remove this overriding once I create a contribution to the upstream container with some flag to skip this method, until then, this custom class will be used.
    
This PR closes yugabyte/yugabyte-db#16929

### DIFF
--- a/src/test/java/io/debezium/connector/yugabytedb/TestHelper.java
+++ b/src/test/java/io/debezium/connector/yugabytedb/TestHelper.java
@@ -425,12 +425,6 @@ public final class TestHelper {
         return container;
     }
 
-    public static String startupCommandString;
-
-    public static String getYBStart() {
-        return startupCommandString;
-    }
-
     public static YugabyteYSQLContainer getYbContainer() {
         return getYbContainer(null, null);
     }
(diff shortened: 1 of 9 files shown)


suranjan (commented):
There should be a test to start a container, create a table, insert, select, restart container, select and verify everything is same.

suranjan:
Looks good.

suranjan (approved):

PR #214, opened 2023-04-25
[yugabyte/yugabyte-db#16946] Handle tablet split scenario with explicit checkpointing

vaibhav-yb:
This PR fixes a bug with explicit checkpoint and tablet splitting wherein let's say a tablet got split and the children got split further, but we haven't received a callback from Kafka about the parent record, in this case no explicit checkpoint was being marked on the `cdc_state` table thus leading us to in a error:

```
The logs from index -1 have been garbage collected and cannot be read
```

The fix to this is to detect that when we receive the tablet split message, then check if we have received the explicit commit callback as well, if we have then handle the split normally, if not then we proceed to add a tablet to a wait list so that we can pause polling on it for a while till the time we do not receive explicit commit callback.

This closes yugabyte/yugabyte-db#16946


suranjan (approved):
LGTM

PR #220, opened 2023-05-26
[yugabyte/yugabyte-db#17551] Make all tests in the connector use explicit checkpointing

vaibhav-yb:
This PR modifies the following things:
1. Converts all tests to use EXPLICIT checkpointing
    a. Added custom engine implementation to use a record committer
    b. Overridden some test functions because the super class members were `private`
2. Fixed an error faced while running tests for snapshot with explicit checkpointing

### DIFF
--- a/src/test/java/io/debezium/connector/yugabytedb/common/TestBaseClass.java
+++ b/src/test/java/io/debezium/connector/yugabytedb/common/TestBaseClass.java
@@ -138,7 +138,7 @@ public class TestBaseClass extends AbstractConnectorTest {
                           DebeziumEngine.CompletionCallback callback) {
     configBuilder
       .with(EmbeddedEngine.ENGINE_NAME, "test-connector")
-      .with(EmbeddedEngine.OFFSET_STORAGE, MemoryOffsetBackingStore.class.getSimpleName())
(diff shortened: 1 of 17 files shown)


src/test/java/io/debezium/connector/yugabytedb/common/TestBaseClass.java
```diff
@@ -92,5 +123,148 @@ protected void restartYugabyteDB(long millisecondsToWait) throws Exception {
...
+
+  public void startEngine(Configuration.Builder configBuilder) {
+    startEngine(configBuilder, (success, message, error) -> {});
+  }
+
+  public void startEngine(Configuration.Builder configBuilder,
+                          DebeziumEngine.CompletionCallback callback) {
+    configBuilder
+      .with(EmbeddedEngine.ENGINE_NAME, "test-connector")
+      .with(StandaloneConfig.OFFSET_STORAGE_FILE_FILENAME_CONFIG, Testing.Files.createTestingFile("file-connector-offsets.txt").getAbsolutePath())
```
suranjan: We should use memory storage

suranjan (changes requested):

suranjan (approved):

PR #223, opened 2023-06-01
[#222] Fix test infra error of not consuming records

### LINKED ISSUE #222: Tests with larger records count failing after changes in test infra
After we changed all the tests to use the explicit checkpoint, the infra caused some issues while running some tests. Upon investigation, it was found that the root cause was the use of `ArrayBlockingQueue` of a fixed size in the `TestBaseClass` which limited our in-memory store of records to a certain configurable number only.

A solution could be not passing the size of the queue while initializing.

vaibhav-yb:
After recent changes where we migrated all the test infra to use explicit checkpointing, it was observed that in the infra was returning inconsistent consumed records intermittently.

Upon investigation, it was discovered that the queue we are using to hold the records can only hold certain number of records at a time which was the culprit. This PR adds a fix by removing the upper limit on the count of records.

Additionally, this closes #222

### DIFF
--- a/src/test/java/io/debezium/connector/yugabytedb/common/TestBaseClass.java
+++ b/src/test/java/io/debezium/connector/yugabytedb/common/TestBaseClass.java
@@ -181,17 +181,15 @@ public class TestBaseClass extends AbstractConnectorTest {
   }
 
   protected int consumeAvailableRecords(Consumer<SourceRecord> recordConsumer) {
-    List<SourceRecord> records = new LinkedList<>();
-
-    // Add all the consumed records to the list.
-    while (!consumedLines.isEmpty()) {
-      records.add(consumedLines.poll());
-    }
-
     if (recordConsumer != null) {
-        records.forEach(recordConsumer);
+        consumedLines.forEach(recordConsumer);
     }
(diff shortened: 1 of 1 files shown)


src/test/java/io/debezium/connector/yugabytedb/common/TestBaseClass.java
```diff
@@ -194,7 +182,12 @@ public void taskStarted() {
 
   protected int consumeAvailableRecords(Consumer<SourceRecord> recordConsumer) {
     List<SourceRecord> records = new LinkedList<>();
-    consumedLines.drainTo(records);
+
```
vrajat, line 185: Nit: Do you need to copy the records to another list? Can `consumedLines` be used directly?

vrajat (commented):
One minor comment. LG in general

vrajat (approved):

src/test/java/io/debezium/connector/yugabytedb/common/TestBaseClass.java
vaibhav-yb: Fixed

PR #229, opened 2023-06-12
[yugabyte/yugabyte-db#17734] Fix loss of precision in timestamp values

vaibhav-yb:
This PR changes the value schema builder and value converter values in the connector by using the proper builders based on the `time.precision.mode`.

Added a test as well to verify the functioning of all 3 temporal modes. Additionally, this closes yugabyte/yugabyte-db#17734


vrajat (approved):

suranjan:
@vaibhav-yb Have we checked the compatibility with the postgres connector?

vaibhav-yb:
@suranjan Yes, the compatibility with Postgres connector has been verified. The behaviour of all three temporal modes is the same now as the one in Postgres connector.

PR #230, opened 2023-06-14
Sync feature branch with main

vaibhav-yb:
Merge main branch changes to feature branch.

PR #231, opened 2023-06-18
Add test to verify atomicity

vaibhav-yb:
This PR adds a simple integration test which verifies that we are atomically streaming the records. At the moment, this PR has a dependency on #230


vaibhav-yb:
1.

PR #233, opened 2023-06-21
[yugabyte/yugabyte-db#17769] Refactoring YugabyteDBOffsetContext

vaibhav-yb:
## Problem
As and when the connector receives the GetChangesResponse from the server, it updates the offset by each record, the default flow looks like:
1. Receive `GetChangesResponse`
2. Iterate over the records and keep updating the offsets
3. After the batch is processed, update the offset with the `OpId` value received in the response

For consistency, the above mentioned flow is altered and we keep the records for sorting first, the flow is effectively:
1. Receive `GetChangesResponse`
2. Add all the records to `Merger`
3. Update the offset to the `OpId` value received in the response in step 1 - note that this is the `from_op_id` for the next `GetChanges` call
4. Process the records from Merger
5. While processing this step, every record also updates the offset for the tablet which is basically overriding the value updated in step 3. This causes an issue where we end up requesting changes from the wrong `from_op_id` and thus receiving a record in the incorrect order.

## Solution
This PR implements the following points by refactoring the `YugabyteDBOffsetContext`:
1. Do not update the `from_op_id` with every record
2. Every record can update a different parameter inside `SourceInfo` in order to decouple it
3. Make sure that we are still storing the correct offsets with/for each record as it will be used in the commitOffset callback in EXPLICIT checkpointing

### Changes part of this pull request
1. Removed unused objects from `YugabyteDBOffsetContext`
2. Added a comment `// Do nothing` while overriding certain functions which we are not using but still need to be overridden
3. Removed unused class `LegacyV1PostgresSourceInfoStructMaker`
4. Removed field `xmin` from all the places because we were anyway passing a `null` value to it everywhere so essentially it was not required and was causing confusion.

This closes yugabyte/yugabyte-db#17769


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
```diff
@@ -216,35 +203,38 @@ public SourceInfo getSourceInfo(YBPartition partition) {
...
+        // TODO: think about making this work
+        return true;
+        // return sourceInfo.isSnapshot();
     }
 
     @Override
     public void preSnapshotStart() {
-        sourceInfo.setSnapshot(SnapshotRecord.TRUE);
-        lastSnapshotRecord = false;
+        // Do nothing.
```
vaibhav-yb, line 213: If I throw `UnsupportedOperationException` here, then the base class which is calling these functions will all start hitting this.

We are only inheriting the methods but not using them to process our snapshot.

vrajat (approved):

PR #234, opened 2023-06-22
[yugabyte/yugabyte-db#17859] Regression with colocated tables

vaibhav-yb:
## Problem

There was a piece of code which was being used to ignore the records for those tables which are not a part of the `table.include.list`. Upon investigation, it was found that there was a logical error wherein if there is a DDL message, then the message was not being ignored and the connector proceeded to register the schema info for that table.

Now when the control flow went into the class `YugabyteDBSchema` - it again checked the filter and saw that the DDL record's table is not included in the filter, and thus it did not register the table schema and further on, when the connector tried to build the schema in `YugabyteDBSchema#buildAndRegisterSchemaForTablet` - we hit the `NullPointerException` as mentioned in the ticket.

## Solution

Remove the check for DDL message from the logic of filtering the records so that we also filter out the DDL message at an early stage rather than pushing it further.

### Test plan
Run the existing tests in `YugabyteDBColocatedTablesTest`

This PR closes yugabyte/yugabyte-db#17859


vrajat (commented):
Does this diff have unrelated changes as well?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -490,8 +490,9 @@ protected void getChanges2(ChangeEventSourceContext context,
 
                             // This is a hack to skip tables in case of colocated tables
                             TableId tempTid = YugabyteDBSchema.parseWithSchema(message.getTable(), pgSchemaNameInRecord);
-                            if (!message.isDDLMessage() && !message.isTransactionalMessage()
+                            if (!message.isTransactionalMessage()
                                   && !new Filters(connectorConfig).tableFilter().isIncluded(tempTid)) {
```
suranjan, line 494: Can we create the filter one time rather than everytime of each message.

src/test/java/io/debezium/connector/yugabytedb/common/TestBaseClass.java
```diff
@@ -183,6 +184,30 @@ public void taskStarted() {
     });
   }
 
+  /**
```
suranjan, line 187: How is this change related?

vrajat (approved):

PR #236, opened 2023-06-27
[yugabyte/yugabyte-db#17854] Bug while transition from snapshot to streaming

vaibhav-yb:
## Problem

It was found out that there is a flow where we set the `tabletSourceInfo` to `0.0` - while snapshotting while streaming snapshot records. This `0.0` came because we do not send any `OpId` with snapshot records but while accessing the records, when we are calling `getTerm` or `getIndex` we are getting the default value `0.0` which is then being used in commit callback as `0.0`

The previous flow also resulted in the error of the form:

```
org.yb.client.CDCErrorException: Server[4ff8d4f19ae24e7aa78c0a43e13bf5ff] INTERNAL_ERROR[code 21]: CDCSDK Trying to fetch already GCed intents for transaction 7efb3374-c9d1-4454-b30e-ddb74baafffe
```

## Solution

A part of the solution was to decouple the offset storage objects which was implemented by #233 - after this, there were other fixes required to resolve issues of invalid checkpoints which are mentioned below and are a part of this PR.
1. Adding a `GetChanges` call with `snapshot_done_key` as the key to mark snapshot completed on service.
2. Initialize the `fromLsn` object upon calling the method `YugabyteDBOffsetContext#initSourceInfo`

### Service diff
https://phorge.dev.yugabyte.com/D26336

### Test plan
The test `YugabyteDBSnapshotTest#snapshotColocatedNonColocatedThenStream` started to fail when we started running tests with explicit checkpointing with our frequent Jenkins runs. The test passes with the changes included in the PR.

This PR also closes yugabyte/yugabyte-db#17854


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -490,7 +490,7 @@ protected void getChanges2(ChangeEventSourceContext context,
 
                             // This is a hack to skip tables in case of colocated tables
                             TableId tempTid = YugabyteDBSchema.parseWithSchema(message.getTable(), pgSchemaNameInRecord);
-                            if (!message.isDDLMessage() && !message.isTransactionalMessage()
+                            if (!message.isTransactionalMessage()
```
vrajat, line 493: this change is from another PR?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -592,20 +616,56 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
...
+                                        Set<String> tabletsWaitingForCallback, YugabyteDBOffsetContext offsetContext) throws Exception {
+      if (this.tabletToExplicitCheckpoint.get(partition.getId()) == null) {
         // If we have no OpId stored in the explicit checkpoint map then that would indicate that
         // we haven't yet received any callback from Kafka even once and we should wait more.
         return;
       }
 
-      if (isSnapshotCompleteMarker(opId)) {
+      OpId opId = OpId.from(this.tabletToExplicitCheckpoint.get(partition.getId()));
+      if (isSnapshotCompleteMarker(opId) && !snapshotCompletedTablets.contains(partition.getId())) {
```
vrajat, line 627: is the 2nd condition required? It is also checked before calling this fn. Maybe an assert is better instead of checking twice? 

Specifically this condition `snapshotCompletedTablets.contains`

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
vaibhav-yb: Yes, fixed after rebase.

vrajat (approved):

PR #242, opened 2023-07-06
Add commit_time from record to Kafka instead of response safe_time

adithya-yb:
While we are pushing records to Kafka, we will consider the commit_time of each record, instead of the response safe_time. 
This also ensure that when we get a Kafka callback acknowledging the record is pushed and persisted in Kafka, we will resuse the same commit_time while setting the checkpoint in cdc_service.


vaibhav-yb (approved):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -545,7 +545,7 @@ protected void getChanges2(ChangeEventSourceContext context,
                                     record.getCdcSdkOpId().getIndex(),
```
vaibhav-yb, line 545: These `getCdcSdkOpId()` need to be changed to `getFromOpId()`

vaibhav-yb (changes requested):

vaibhav-yb (approved):

PR #246, opened 2023-07-11
Explicit checkpoint wait on Tablet Split is now based on LastRecordCheckpoint

adithya-yb:
We will populate the last seen record's checkpoint in the "lastRecordCheckpoint" field. This is meant to have the explicit checkpoint details of the last seen record from any response (i,e the from_op_id and the record’s commit_time). We will try to update this value in every GetChanges call.

After “lastRecordCheckpoint” is updated as needed, we will then see the response checkpoint from the current GetChanges response. If the response checkpoint is higher :

1. We will store the response’s checkpoint, if the response’s checkpoint is higher than “lastReadRecordCheckpoint”
2. We will add this tablet to a wait-list, where we will first wait until the ExplicitCheckpoint of the tablet is equal to: “lastRecordCheckpoint”.
3. After the ExplicitCheckpoint was equal to “lastRecordCheckpoint”, then we can update the ExplicitCheckpoint value of the tablet to the response’s checkpoint i.e directly update ExplicitCheckpoint in the “tabletToExplicitCheckpoint“ map.

This will take care of updating the ExplicitCheckpoint in cases of NoOp messages in the RAFT WAL , and also take care of tablets splitting right after a previous split in which case there will be no records to checkpoint.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
@@ -396,16 +396,16 @@ public class YugabyteDBStreamingChangeEventSource implements
                         // checkpoint is the same as from_op_id, if yes then handle the tablet
                         // for split.
                         CdcSdkCheckpoint explicitCheckpoint = tabletToExplicitCheckpoint.get(part.getId());
-                        OpId lastCommitLsn = offsetContext.getTabletSourceInfo().get(part.getId()).lastCommitLsn();
+                        OpId lastRecordCheckpoint = offsetContext.getSourceInfo(part).lastRecordCheckpoint();
 
-                        if (explicitCheckpoint != null && (lastCommitLsn == null || lastCommitLsn.isLesserThanOrEqualTo(explicitCheckpoint))) {
+                        if (explicitCheckpoint != null && (lastRecordCheckpoint == null || lastRecordCheckpoint.isLesserThanOrEqualTo(explicitCheckpoint))) {
                             // At this position, we know we have received a callback for split tablet
                             // handle tablet split and delete the tablet from the waiting list.
 
                             // Call getChanges to make sure checkpoint is set on the cdc_state table.
                             LOGGER.info("Setting explicit checkpoint is set to {}.{}", explicitCheckpoint.getTerm(), explicitCheckpoint.getIndex());
                             setCheckpointWithGetChanges(tableIdToTable.get(part.getTableId()), part,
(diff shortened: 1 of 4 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -397,14 +396,16 @@ protected void getChanges2(ChangeEventSourceContext context,
...
+
+                        if (explicitCheckpoint != null && (lastRecordCheckpoint == null || lastRecordCheckpoint.isLesserThanOrEqualTo(explicitCheckpoint))) {
                             // At this position, we know we have received a callback for split tablet
                             // handle tablet split and delete the tablet from the waiting list.
 
                             // Call getChanges to make sure checkpoint is set on the cdc_state table.
                             LOGGER.info("Setting explicit checkpoint is set to {}.{}", explicitCheckpoint.getTerm(), explicitCheckpoint.getIndex());
                             setCheckpointWithGetChanges(tableIdToTable.get(part.getTableId()), part,
-                                cp, explicitCheckpoint, schemaNeeded.get(part.getId()),
+                                lastRecordCheckpoint, explicitCheckpoint, schemaNeeded.get(part.getId()),
```
vaibhav-yb, line 408: This will call the `GetChanges` API by setting the `from_op_id` as `lastRecordCheckpoint`, are we sure that won't have any side effects?

vaibhav-yb (commented):
The changes seem fine, I am a little skeptical about the comment I provided, please confirm once if that will not have unintended effects.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
adithya-yb: Yes, need to revert this to cp again

vaibhav-yb (changes requested):
Two things, in the `YugabyteDBStreamingChangeEventSource` class:

1. In the `handleTabletSplit` method, add a line to remove the instance of split tablet from the explicit map:

```
// Remove the corresponding entry for explicit checkpoint as well.
tabletToExplicitCheckpoint.remove(splitTabletId);
```

2. Initialise the explicit checkpoint for children tablets in the `addTabletIfNotPresent`:
```
tabletToExplicitCheckpoint.put(p.getId(), OpId.from(pair.getCdcSdkCheckpoint()).toCdcSdkCheckpoint());
```

vaibhav-yb (approved):

PR #250, opened 2023-07-19
Add `commit_time` to the checkpoint in case of consistent streaming

vaibhav-yb:


suranjan (approved):
LGTM

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
```diff
@@ -319,11 +320,19 @@ private void dispatchMessage(YugabyteDBOffsetContext offsetContext, Map<String,
             return;
         }
 
+        long requiredCommitTime = record.getRowMessage().getCommitTime();
+        if (message.getOperation() != Operation.COMMIT) {
+            // Only when we see a 'COMMIT' record, can we set the checkpoint
+            // with the same time as the current transaction's commit time, otherwise we need to replay this
+            // transaction in cases of connector restart, hence we will set the checkpoint to be the commit_time - 1.
+            requiredCommitTime -= 1;
```
adithya-yb, line 328: @vaibhav-yb  We always have to use 
record.getRowMessage().getCommitTime() - 1 , even in the case of a COMMIT record 

Otherwise there can be data loss, in scenarios where two txns share the commit_time

suranjan:
Need to make change to subtract 1 in all cases.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
adithya-yb: This has to be done on both the files

adithya-yb (approved):

PR #254, opened 2023-07-25
[yugabyte/yugabyte-db#18245] Use `lastRecordCheckpoint` to detect empty snapshot

vaibhav-yb:
## Problem

Today, in order to check if a tablet has empty snapshot, we just rely on the response size i.e. `GetChangesResponse` will have an empty record list, or in other terms a list of size 0. However, this logic is not enough for all the conditions where we genuinely can have empty responses, for example:
1. The first `GetChanges` call in cases of snapshot - this is only used to bootstrap snapshot on service.
2. Empty responses because of network or intermittent service issues.

## Solution

This PR adds the following changes to fix the problem described:
1. Adds a getter in `SourceInfo` to check if we haven't seen any record yet.
3. Disallows waiting for callback if the tablet has no data to snapshot, this is checked if the following conditions are met
    a. The `OpId` in `GetChangesResponse` is the snapshot complete marker.
    b. We haven't seen any record in snapshot (checked using the getter method)

### Misc changes

1. Upgraded `yb-client` version.
2. Changed a log to print the explicit checkpoint value as well.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -385,6 +385,11 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
                   continue;
                 }
 
+                CdcSdkCheckpoint explicitCdcSdkCheckpoint = null;
+                if (taskContext.shouldEnableExplicitCheckpointing()) {
+                  explicitCdcSdkCheckpoint = tabletToExplicitCheckpoint.get(part.getId());
+                }
+
                 OpId cp = previousOffset.snapshotLSN(part);
 
                 if (LOGGER.isDebugEnabled()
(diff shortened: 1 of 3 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -495,6 +493,23 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
                                           resp.getWriteId(), resp.getSnapshotTime());
                 LOGGER.debug("Final OpId for tablet {} is {}", part.getId(), finalOpId);
 
+                // In cases where the tablet is empty, the response checkpoint can still move ahead and we should
+                // also move the explicit checkpoint forward, given that it was already greater than the lsn of the last seen valid record.
+                // Otherwise, we will be stuck waiting for callback on an empty tablet.
+                if (taskContext.shouldEnableExplicitCheckpointing()) {
+                  // If the response doesn't have any record and we got the snapshot end marker, we know the snapshot.
```
adithya-yb, line 500: looks life we forgot to complete the sentence here.

adithya-yb (approved):

PR #256, opened 2023-08-04
[yugabyte/yugabyte-db#18542] Fix NPE and incorrect reinitialisation of set

vaibhav-yb:
## Problem

1. There is a log line where we are accessing the `toString()` method on an object which can be null which in turn is throwing a `NullPointerException` with a similar stacktrace as

```
Caused by: java.lang.NullPointerException
         at io.debezium.connector.yugabytedb.YugabyteDBSnapshotChangeEventSource.doExecute(YugabyteDBSnapshotChangeEventSource.java:398)
         at io.debezium.connector.yugabytedb.YugabyteDBSnapshotChangeEventSource.execute(YugabyteDBSnapshotChangeEventSource.java:136)
         ... 9 more

```

2. Another issue is the reinitialisation of a set being used to store the list of tablets which should wait for callbacks. Over here, we have a set which is getting reinitialised in every iteration of the loop so the flow is something like:

```
Start Loop:
1. Initialise set
2. Add tablet list to the set
3. Upon loop reiteration, reinitialise - this step is causing a loss of the previous state.
```

So let's say in the previous state, the connector knew it has to wait for callback on some tablet and not call further `GetChanges` until it receives a callback, the state is now gone and the connector will end up calling `GetChanges` with a checkpoint not meant for snapshot - thus in those cases it will also see a `SAFEPOINT` record which shouldn't have been the case as pointed by the ticket yugabyte/yugabyte-db#18464

## Solution
1. Instead of accessing the method `toString()` add the object to the log line directly, this way, Java will itself understand whether to print `null` or call the `toString()` method.
2. Move the initialisation of the set outside of the loop.

### Test Plan

Modified a test by adding a wait time before the commit callback to reproduce the error.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -392,10 +392,10 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
 
                 OpId cp = previousOffset.snapshotLSN(part);
 
-                if (LOGGER.isDebugEnabled()
+                if (true || LOGGER.isDebugEnabled()
                     || (connectorConfig.logGetChanges() && System.currentTimeMillis() >= (lastLoggedTimeForGetChanges + connectorConfig.logGetChangesIntervalMs()))) {
                   LOGGER.info("Requesting changes for tablet {} from OpId {} for table {} with explicit checkpoint {}",
-                              tabletId, cp, table.getName(), explicitCdcSdkCheckpoint.toString());
+                              tabletId, cp, table.getName(), explicitCdcSdkCheckpoint);
                   lastLoggedTimeForGetChanges = System.currentTimeMillis();
                 }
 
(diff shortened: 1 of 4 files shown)


adithya-yb (approved):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -98,6 +103,7 @@ public YugabyteDBSnapshotChangeEventSource(YugabyteDBConnectorConfig connectorCo
         this.tabletSafeTime = new HashMap<>();
         this.tableIdToTable = new HashMap<>();
         this.shouldWaitForCallback = new HashMap<>();
+        this.tabletsWaitingForCallback = new HashSet<>();
```
suranjan, line 106: Where do we add/remove from this set?

vaibhav-yb: Add here once we detect that this tablet hasn't caught up on the explicit checkpoint: https://github.com/yugabyte/debezium-connector-yugabytedb/blob/main/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java#L559

Remove once we receive the callback and know that the tablet has completed the snapshot: https://github.com/yugabyte/debezium-connector-yugabytedb/blob/main/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java#L646

PR #257, opened 2023-08-07
[yugabyte/yugabyte-db#18542] Fix data loss caused by incorrect check for no records in batch

vaibhav-yb:
## Problem

There are cases where snapshots can be empty i.e. genuine cases where there is nothing to snapshot. In those cases we use a helper function to determine whether the partition/tablet has seen any record in the connector or not.

The check is simple:

```java
public boolean noRecordSeen() {
  return lastRecordCheckpoint == null;
}
```

The above check is based on the assumption that `lastRecordCheckpoint` will remain uninitialised if there is no record yet. However, in some cases, it is possible that we initialise this object with `YugabyteDBOffsetContext#snapshotStartLsn` or `YugabyteDBOffsetContext#streamingStartLsn` as referenced here: https://github.com/yugabyte/debezium-connector-yugabytedb/blob/main/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java#L261

Thus, the above logic would fail.

## Solution

This PR adds the solution by modifying the logic for the same and also comparing the `lastRecordCheckpoint` with the `snapshotStartLsn` and `streamingStartLsn` effectively meaning that the connector will consider the snapshot batch as empty if any of the following conditions are met (checked in the given order only):
1. `lastRecordCheckpoint` is null
2. `lastRecordCheckpoint` is equal to either `snapshotStartLsn` or `streamingStartLsn`.

### Test plan
Tested using a failing stress test which was reproducing this issue: `test_cdc_delete_stream_id_with_connector`

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -386,17 +386,12 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
                   continue;
                 }
 
-                CdcSdkCheckpoint explicitCdcSdkCheckpoint = null;
-                if (taskContext.shouldEnableExplicitCheckpointing()) {
-                  explicitCdcSdkCheckpoint = tabletToExplicitCheckpoint.get(part.getId());
-                }
-
                 OpId cp = previousOffset.snapshotLSN(part);
 
                 if (LOGGER.isDebugEnabled()
                     || (connectorConfig.logGetChanges() && System.currentTimeMillis() >= (lastLoggedTimeForGetChanges + connectorConfig.logGetChangesIntervalMs()))) {
-                  LOGGER.info("Requesting changes for tablet {} from OpId {} for table {} with explicit checkpoint {}",
-                              tabletId, cp, table.getName(), explicitCdcSdkCheckpoint);
+                  LOGGER.info("Requesting changes for tablet {} from OpId {} for table {}",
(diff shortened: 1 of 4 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -494,21 +494,11 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
...
-
-                  OpId lastRecordCheckpoint = sourceInfo.lastRecordCheckpoint();
-                  if (sourceInfo.noRecordSeen() || lastRecordCheckpoint.isLesserThanOrEqualTo(explicitCdcSdkCheckpoint)) {
-                    tabletToExplicitCheckpoint.put(part.getId(), finalOpId.toCdcSdkCheckpoint());
-                  }
+                // If the response doesn't have any record, it is safe to assume that we should not wait
+                // for the callback to come and that we can proceed further in processing this particular tablet.
+                if (previousOffset.getSourceInfo(part).noRecordSeen()) {
+                  LOGGER.info("Should not wait for callback on tablet {}", part.getId());
+                  shouldWaitForCallback.put(part.getId(), Boolean.FALSE);
```
suranjan, line 501: shouldWaitForCallback is map with Boolean value? Shouldn't we use Set for it?

vaibhav-yb: We can use. Previously it was used in other contexts as well, I will change this to use set.

suranjan (approved):

PR #265, opened 2023-09-11
[yugabyte/yugabyte-db#12751] Added support for multi node fail over connection

Sumukh-Phalgaonkar:
## Problem
Currently, the Debezium Connector for YugabyteDB takes in a `database.hostname` which is the IP of a node where the connector performs a one time operation to validate the user authentication.

This can create a problem as the current setup leads to a hard dependency on the provided node i.e. in case the connector restarts and the DB node crashes at that time then the connector will not be able to validate the connection and as a result of this, it will not proceed forward and fail at the validation step itself.

## Solution
With this PR `database.hostname` config can now take in a comma separated addresses (ex `ip1:port1,ip2:port2, ip3:port3`) of multiple DB nodes. In case a node isn't available at the time of validation, the connector tries to connect to the next node. 
The earlier way of specifying a single IP of a node is also supported.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
@@ -222,14 +222,16 @@ public class YugabyteDBStreamingChangeEventSource implements
             }
         }
 
-        short retryCountForBootstrapping = 0;
+        short maxBootstrapRetries = 25;
         for (Pair<String, String> entry : tabletPairList) {
             // entry is a Pair<tableId, tabletId>
             boolean shouldRetry = true;
-            while (retryCountForBootstrapping <= connectorConfig.maxConnectorRetries() && shouldRetry) {
+            short retryCountForBootstrapping = 0;
+            while (retryCountForBootstrapping <= maxBootstrapRetries && shouldRetry) {
                 try {
                     if (!tabletsWithoutBootstrap.contains(entry.getValue())) {
-                        bootstrapTablet(this.syncClient.openTableByUUID(entry.getKey()), entry.getValue());
+                        YBTable table = this.syncClient.openTableByUUID(entry.getKey());
+                        bootstrapTablet(table, entry.getValue());
                     } else {
                         LOGGER.info("Skipping bootstrap for table {} tablet {} as it has a checkpoint", entry.getKey(), entry.getValue());
                     }
@@ -242,14 +244,14 @@ public class YugabyteDBStreamingChangeEventSource implements
                     // The connector should go for a retry if any exception is thrown
                     shouldRetry = true;
 
-                    if (retryCountForBootstrapping > connectorConfig.maxConnectorRetries()) {
-                        LOGGER.error("Failed to bootstrap the tablet {} after {} retries", entry.getValue(), connectorConfig.maxConnectorRetries());
+                    if (retryCountForBootstrapping > maxBootstrapRetries) {
+                        LOGGER.error("Failed to bootstrap the tablet {} after {} retries", entry.getValue(), maxBootstrapRetries);
                         throw e;
                     }
 
                     // If there are retries left, perform them after the specified delay.
                     LOGGER.warn("Error while trying to bootstrap tablet {}; will attempt retry {} of {} after {} milli-seconds. Exception message: {}",
-                            entry.getValue(), retryCountForBootstrapping, connectorConfig.maxConnectorRetries(), connectorConfig.connectorRetryDelayMs(), e.getMessage());
+                            entry.getValue(), retryCountForBootstrapping, maxBootstrapRetries, connectorConfig.connectorRetryDelayMs(), e.getMessage());
 
                     try {
                         final Metronome retryMetronome = Metronome.parker(Duration.ofMillis(connectorConfig.connectorRetryDelayMs()), Clock.SYSTEM);
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
@@ -222,6 +222,10 @@ public class YugabyteDBStreamingChangeEventSource implements
             }
         }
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/debezium/connector/yugabytedb/connection/YugabyteDBConnection.java
```diff
@@ -138,7 +171,12 @@ static JdbcConfiguration addDefaultSettings(JdbcConfiguration configuration, Str
      * @return a {@code String} where the variables in {@code urlPattern} are replaced with values from the configuration
      */
     public String connectionString() {
-        return connectionString(URL_PATTERN);
+        String hostName = config.getHostname();
+        if(hostName.contains(":")){
```
suranjan, line 175: correct formatting

suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -222,14 +222,16 @@ protected void bootstrapTabletWithRetry(List<Pair<String,String>> tabletPairList
             }
         }
 
-        short retryCountForBootstrapping = 0;
+        int maxBootstrapRetries = connectorConfig.maxConnectorRetries() * 5;
```
suranjan, line 225: Put some comment on why are we doing this.

suranjan (changes requested):
Put also a negative test with wrong URL pattern and see what error we see.

src/main/java/io/debezium/connector/yugabytedb/connection/YugabyteDBConnection.java
```diff
@@ -61,15 +61,17 @@ public class YugabyteDBConnection extends JdbcConnection {
 
     private static Logger LOGGER = LoggerFactory.getLogger(YugabyteDBConnection.class);
 
-    private static final String URL_PATTERN = "jdbc:postgresql://${" + JdbcConfiguration.HOSTNAME + "}:${"
+    private static final String MULTI_HOST_URL_PATTERN = "jdbc:postgresql://${" + JdbcConfiguration.HOSTNAME + "}/${" + JdbcConfiguration.DATABASE + "}";
```
vaibhav-yb, line 64: I think we should be using our smart driver rather than the one from Postgres.

src/main/java/io/debezium/connector/yugabytedb/connection/YugabyteDBConnection.java
```diff
@@ -114,6 +134,19 @@ public YugabyteDBConnection(YugabyteDBConnectorConfig config,
         }
     }
 
+    public YugabyteDBConnection(YugabyteDBConnectorConfig config,
+                                YugabyteDBTypeRegistry yugabyteDBTypeRegistry, String connectionUsage) {
+        this(config,yugabyteDBTypeRegistry, connectionUsage, config.getJdbcConfig().getHostname().contains(":")
```
vaibhav-yb, line 139: For the code snippet which uses a conditional operator here to select which URL pattern to use, we can create some getter in `YugabyteDBConnectorConfig` itself, for example:
```java
public <connectionFactoryType> getConnectionFactory() {
return getHostname().contains(":")
                        ? JdbcConnection.patternBasedFactory(MULTI_HOST_URL_PATTERN,
                            org.postgresql.Driver.class.getName(),
                            YugabyteDBConnection.class.getClassLoader(),
                            JdbcConfiguration.PORT.withDefault(PORT.defaultValueAsString()))
                        : JdbcConnection.patternBasedFactory(SINGLE_HOST_URL_PATTERN,
                                org.postgresql.Driver.class.getName(),
                                YugabyteDBConnection.class.getClassLoader(), JdbcConfiguration.PORT
                                        .withDefault(PORT.defaultValueAsString()))
}
```

And then replace all the instances of this conditional operator with `config.getConnectionFactory()`

vaibhav-yb (changes requested):

Sumukh-Phalgaonkar:
> Put also a negative test with wrong URL pattern and see what error we see.

@suranjan I didn't get this. We don't let the user directly decide the URL pattern to use. If the `hostname` contains just the IP then the older URL pattern will be used. If it contains both IP and port then newer one will be used. The user can only specify the value of `hostname` and appropriate URL pattern will be used.

If the user specifies wrong address for `hostname` then an error saying ```PSQL connection refused to <address> Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.``` is thrown.

src/main/java/io/debezium/connector/yugabytedb/connection/YugabyteDBConnection.java
Sumukh-Phalgaonkar: IIRC we had decided to drop smart driver because it needed a lot of changes to be made. @suranjan  should we use smart driver or can we continue with postgres one?

suranjan (approved):

suranjan:
> > Put also a negative test with wrong URL pattern and see what error we see.
> 
> @suranjan I didn't get this. We don't let the user directly decide the URL pattern to use. If the `hostname` contains just the IP then the older URL pattern will be used. If it contains both IP and port then newer one will be used. The user can only specify the value of `hostname` and appropriate URL pattern will be used.
> 
> If the user specifies wrong address for `hostname` then an error saying `PSQL connection refused to <address> Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.` is thrown.

This is fine. Did we also check with string:string:string kind of URL?

Sumukh-Phalgaonkar:
> > > Put also a negative test with wrong URL pattern and see what error we see.
> > 
> > 
> > @suranjan I didn't get this. We don't let the user directly decide the URL pattern to use. If the `hostname` contains just the IP then the older URL pattern will be used. If it contains both IP and port then newer one will be used. The user can only specify the value of `hostname` and appropriate URL pattern will be used.
> > If the user specifies wrong address for `hostname` then an error saying `PSQL connection refused to <address> Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.` is thrown.
> 
> This is fine. Did we also check with string:string:string kind of URL?

Yes it gives PSQLException saying connection attempt failed

vaibhav-yb (approved):

PR #270, opened 2023-09-25
Fix for connector losing tablet list upon restart and RPC failures.

vaibhav-yb:
## Problem

1. If the connector restarts just one task, and if any tablet has been split, then chances are that we lose track of the tablet list after restart and we might start getting `RuntimeException` with a log message of the workaround saying that restarting the connector would solve it.
2. Intermittently, we get a `tablet=null` error while calling RPCs.

## Solution

1. Now if we see that any tablet is missing from the list, we simply call the API `GetTabletListToPollForCDC` again to get its children and we add them to the list. By moving this way, we recreate the proper tablet list to be polled for.
2. If we get a `tablet=null` error, we recreate `YBClient` and move ahead.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -55,7 +55,7 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
     private final YugabyteDBConnection connection;
 
     private final AsyncYBClient asyncClient;
-    private final YBClient syncClient;
+    private YBClient syncClient;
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -592,6 +592,12 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
                        this.connectorConfig.maxConnectorRetries(), 
                        this.connectorConfig.connectorRetryDelayMs(), e);
 
+          if (e.toString().contains("tablet=null")) {
```
suranjan, line 595: Can't YBClient throw a different error in case of tablet=null?

suranjan (changes requested):

PR #272, opened 2023-09-26
[yugabyte/yugabyte-db#19293] Modify logic to decide whether snapshot is complete

vaibhav-yb:
## Problem

The first GetChanges call in snapshot only tells the service to mark the snapshot boundary i.e. the snapshot is bootstrapped.  When snapshot is bootstrapped, the service marks certain `OpId` value in the `cdc_state` table to indicate snapshot end boundary.

Now if the connector fails just after the bootstrap GetChanges call and is restarted, it executes a `GetCheckpoint` RPC call to decide whether or not to take a snapshot, before this PR, the logic was solely based on the presence of a non-empty snapshot key and invalid checkpoint as follows:
1. If there is a non-empty snapshot key then snapshot is not completed and we need to continue taking snapshot, otherwise check for condition 2.
2. If the `OpId` is invalid i.e. term=-1 and index = -1 then take snapshot, otherwise the assumption is that the snapshot is complete.

However, the condition 2 was also true for cases where a snapshot was just bootstrapped and the connector skips the snapshot this time and proceeds to streaming.

## Solution

The fix is to change the logic which we use to decide whether or not to take snapshot:
1. If snapshot key is present in the response (`snapshot_key != null`) - we need to proceed for snapshot.
2. If it is not present (`snaphshot_key == null`), then check if it is invalid, if yes, then take snapshot, otherwise we know that the tablet is already in the streaming mode.

Additionally, the flow has been refactored a little to increase readability, changes include:
1. `hasSnapshotCompletedPreviously` now only takes the `GetCheckpointResponse` called previously - effectively reducing the `GetCheckpoint` RPC calls to half
2. The complete evaluation logic whether to snapshot completed earlier or we need to take snapshot after calling `SetCheckpoint` RPC has been refactored to a method `isSnapshotRequired`
3. Effectively, the flow is now:
    a. Filter tables/tablets based on `snapshot.include.collection.list`
    b. Iterate over the complete tablet list
        i. If table is in the filtered list then check if we need to take snapshot using the method `isSnapshotRequired`
        ii. If table is not in the filtered list, then simply bootstrap the tablet for streaming
    c. Start polling for snapshot on the tablet list for snapshot

### Related service diff

https://phorge.dev.yugabyte.com/D28814

### Test Plan

Added tests:
1. YugabyteDBSnapshotTest#shouldSnapshotWithFailureAfterBootstrapSnapshotCall
2. YugabyteDBSnapshotTest#shouldSnapshotWithFailureAfterSettingInitialCheckpoint
3. YugabyteDBSnapshotTest#shouldNotSnapshotAgainIfSnapshotCompletedOnce
4. YugabyteDBSnapshotTest#shouldContinueStreamingInNeverAfterSnapshotCompleteInInitialOnly

To run the tests, following command can be used:

```
mvn clean test -Dtest=YugabyteDBSnapshotTest#shouldSnapshotWithFailureAfterBootstrapSnapshotCall

mvn clean test -Dtest=YugabyteDBSnapshotTest#shouldSnapshotWithFailureAfterSettingInitialCheckpoint

mvn clean test -Dtest=YugabyteDBSnapshotTest#shouldNotSnapshotAgainIfSnapshotCompletedOnce

mvn clean test -Dtest=YugabyteDBSnapshotTest#shouldContinueStreamingInNeverAfterSnapshotCompleteInInitialOnly
```

This PR closes yugabyte/yugabyte-db#19293

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
         <version.kafka>3.3.1</version.kafka>
         <version.org.slf4j>1.7.36</version.org.slf4j>
         <version.logback>1.4.0</version.logback>
-        <version.ybclient>0.8.62-20230901.095631-3</version.ybclient>
+        <version.ybclient>0.8.64-SNAPSHOT</version.ybclient>
         <version.gson>2.8.9</version.gson>
 
         <!--
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
         <version.kafka>3.3.1</version.kafka>
         <version.org.slf4j>1.7.36</version.org.slf4j>
         <version.logback>1.4.0</version.logback>
-        <version.ybclient>0.8.64-SNAPSHOT</version.ybclient>
+        <version.ybclient>0.8.64-20230929.065833-2</version.ybclient>
         <version.gson>2.8.9</version.gson>
 
         <!--
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -45,6 +45,7 @@ import io.debezium.util.Strings;
 
 public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeEventSource<YBPartition, YugabyteDBOffsetContext> {
     private static final Logger LOGGER = LoggerFactory.getLogger(YugabyteDBSnapshotChangeEventSource.class);
+    public static boolean FAIL_AFTER_BOOTSTRAP_GET_CHANGES;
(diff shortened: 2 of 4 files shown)


suranjan (approved):

pom.xml
```diff
@@ -47,7 +47,7 @@
         <version.kafka>3.3.1</version.kafka>
         <version.org.slf4j>1.7.36</version.org.slf4j>
         <version.logback>1.4.0</version.logback>
-        <version.ybclient>0.8.62-20230901.095631-3</version.ybclient>
+        <version.ybclient>0.8.64-SNAPSHOT</version.ybclient>
```
vaibhav-yb, line 50: Note to self: This needs to be updated to correct yb-client version once service diff is merged.

suranjan (approved):

PR #273, opened 2023-10-04
[yugabyte/yugabyte-db#19394] Snapshot should be resumed from `snapshot_key`

vaibhav-yb:
## Problem

The PR https://github.com/yugabyte/debezium-connector-yugabytedb/pull/272 has introduced one regression in the connector with regard to the snapshot resume capability. This was discovered when one of the connector unit tests started failing i.e. `YugabyteDBSnapshotResumeTest`

Logic before this diff:

1. Get the checkpoint on a tablet using `GetCheckpoint` RPC
2. If a `snapshot_key` is present, then take the snapshot
3. Start snapshot process from `YugabyteDBOffsetContext#snapshotStartLsn`

Now, in step 3 above, we are assigning a `from_op_id` which indicates to take snapshot from the beginning regardless of whether the `GetCheckpoint` returns a checkpoint to resume.

## Solution

This diff modifies the `YugabyteDBSnapshotChangeEventSource` and uses the `GetCheckpointResponse` to evaluate what is the checkpoint we should be using as `from_op_id` while making the first `GetChanges` call for snapshot.

Note that a method `getSnapshotStartLsn(GetCheckpointResponse)` is introduced which is only called when we have successfully evaluated that we need to take a snapshot. The method uses the following logic:

1. If the term/index is valid and non-zero then we should directly use the checkpoint from the `GetCheckpointResponse`
    a. We are saying non-zero here because if the connector crashes just after setting the checkpoint to mark the stream as active and then restarts, we would get a checkpoint with `0.0` value which again indicates that we need to start the snapshot from the beginning
2. In any other case, we know we have to start from the beginning so we start from `YugabyteDBOffsetContext#snapshotStartLsn`

### Service diff

https://phorge.dev.yugabyte.com/D29060

### Test plan

```
mvn clean test -Dtest=YugabyteDBSnapshotResumeTest#verifySnapshotIsResumedFromKey
```

This PR closes yugabyte/yugabyte-db#19394

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
@@ -206,7 +206,7 @@ public class YugabyteDBConsistentStreamingSource extends YugabyteDBStreamingChan
                                         cdcException.printStackTrace();
                                     }
 
-                                    handleTabletSplit(cdcException, tabletPairList, offsetContext, streamId, schemaNeeded);
+                                    handleTabletSplit(part.getTabletId(), tabletPairList, offsetContext, streamId, schemaNeeded);
 
                                     // Break out of the loop so that the iteration can start afresh on the modified list.
                                     break;
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
@@ -192,7 +192,7 @@ public class YugabyteDBConsistentStreamingSource extends YugabyteDBStreamingChan
                         if (merger.isSlotEmpty(tabletId)) {
                             try {
                                 response = this.syncClient.getChangesCDCSDK(
-                                        table, streamId, tabletId, cp.getTerm(), cp.getIndex(), cp.getKey(),
+                                        table, streamId, tabletId, cp.getTerm(), cp.getIndex(), (cp.getKey() == null) ? "".getBytes() : cp.getKey(),
                                         cp.getWrite_id(), cp.getTime(), schemaNeeded.get(tabletId),
                                         taskContext.shouldEnableExplicitCheckpointing() ? tabletToExplicitCheckpoint.get(part.getId()) : null,
                                         tabletSafeTime.getOrDefault(part.getId(), -1L), offsetContext.getWalSegmentIndex(part));
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
@@ -192,7 +192,7 @@ public class YugabyteDBConsistentStreamingSource extends YugabyteDBStreamingChan
                         if (merger.isSlotEmpty(tabletId)) {
                             try {
                                 response = this.syncClient.getChangesCDCSDK(
-                                        table, streamId, tabletId, cp.getTerm(), cp.getIndex(), (cp.getKey() == null) ? "".getBytes() : cp.getKey(),
+                                        table, streamId, tabletId, cp.getTerm(), cp.getIndex(), cp.getKey(),
                                         cp.getWrite_id(), cp.getTime(), schemaNeeded.get(tabletId),
                                         taskContext.shouldEnableExplicitCheckpointing() ? tabletToExplicitCheckpoint.get(part.getId()) : null,
                                         tabletSafeTime.getOrDefault(part.getId(), -1L), offsetContext.getWalSegmentIndex(part));
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -49,6 +49,8 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
     // Test only member variables, DO NOT try to modify in the source code.
     public static boolean FAIL_AFTER_BOOTSTRAP_GET_CHANGES;
     public static boolean FAIL_AFTER_SETTING_INITIAL_CHECKPOINT;
+    public static boolean TRACK_EXPLICIT_CHECKPOINTS = false;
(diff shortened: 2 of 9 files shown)


src/main/java/io/debezium/connector/yugabytedb/connection/OpId.java
```diff
@@ -85,7 +85,7 @@ public String toString() {
         return "" +
                 "term=" + term +
                 ", index=" + index +
-                ", key=" + Arrays.toString(key) +
+                ", key=" + ((key == null) ? "" : Arrays.toString(key)) +
```
asrinivasanyb, line 88: Distinguish between NULL and empty values of the key even while printing

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotResumeTest.java
```diff
@@ -89,51 +88,23 @@ public void verifySnapshotIsResumedFromKey() throws Exception {
...
+		for (int i = 0; i < 10; ++i) {
+			remainingConsumedRecords += consumeAvailableRecords(record -> {});
+			TestHelper.waitFor(Duration.ofSeconds(2));
 		}
 
-		assertEquals(recordsCount, totalConsumedRecords.get());
+		LOGGER.info("Remaining consumed record count: {}", remainingConsumedRecords);
+		assertNotEquals(finalTotalConsumedSoFar, remainingConsumedRecords);
+
+		assertEquals(recordsCount - finalTotalConsumedSoFar, records.size());
```
asrinivasanyb, line 107: To confirm that explicitCheckpointing works we need to assert that the "remaining consumed records" is equal to an appropriate value (100 ? size of.2 batches)

src/main/java/io/debezium/connector/yugabytedb/connection/OpId.java
```diff
@@ -74,7 +74,7 @@ public static OpId valueOf(String stringId) {
      * </pre>
      */
     public String toSerString() {
-        String keyStr = Base64.getEncoder().encodeToString(key);
+        String keyStr = (key == null) ? "" : Base64.getEncoder().encodeToString(key);
```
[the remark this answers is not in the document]
vaibhav-yb: Added the distinction in the `toString()` method. `toSerString()` is only used to store checkpoint values as offsets.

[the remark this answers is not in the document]
asrinivasanyb: This distinction between NULL value (semantics being NOT PRESENT) vs empty string (PRESENT and empty string value) is important in all contexts because these two represent different things. I think, in this context of checkpoints, it is all the more important to capture this distinction

suranjan (approved):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
```diff
@@ -192,7 +192,7 @@ protected void getChanges2(ChangeEventSourceContext context, YBPartition ybParti
                         if (merger.isSlotEmpty(tabletId)) {
                             try {
                                 response = this.syncClient.getChangesCDCSDK(
-                                        table, streamId, tabletId, cp.getTerm(), cp.getIndex(), cp.getKey(),
+                                        table, streamId, tabletId, cp.getTerm(), cp.getIndex(), (cp.getKey() == null) ? "".getBytes() : cp.getKey(),
```
suranjan, line 195: will we be passing empty or null?

suranjan (changes requested):

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotTest.java
```diff
@@ -375,7 +375,7 @@ public void snapshotColocatedNonColocatedThenStream(boolean initialOnly) throws
     }
 
     @ParameterizedTest
-    @ValueSource(booleans = {true, false})
+    @ValueSource(booleans = {false})
```
suranjan: Does this test break colocation tests?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
vaibhav-yb: Now we will be passing null, the handling will be done in the `GetCheckpointRequest.java` in the yb-client.

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotTest.java
vaibhav-yb: This was a test only change, I was running couple of them to verify. I have reverted this too.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -346,16 +348,38 @@ protected boolean isSnapshotRequired(GetCheckpointResponse getCheckpointResponse
...
 
+    /**
+     * This method will ONLY be called when we are sure that we have to take the snapshot.
+     * @param getCheckpointResponse
+     * @return the from_op_id to start snapshot with
+     */
+    protected OpId getSnapshotStartLsn(GetCheckpointResponse getCheckpointResponse) {
+      Objects.requireNonNull(getCheckpointResponse);
+
+      if (getCheckpointResponse.getTerm() > 0 && getCheckpointResponse.getIndex() > 0) {
```
suranjan, line 372: Can the came condition of valid lsn be used here? Or do we see any issue if OpId is (0,0)? We have to be careful.

suranjan (approved):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
vaibhav-yb: It can be `0.0` where the connector just failed after setting the checkpoint for the first time i.e. marking the stream as active - but even in that case, we only want the snapshot to be requested from `-1.-1` as this is still the beginning.

src/main/java/io/debezium/connector/yugabytedb/connection/OpId.java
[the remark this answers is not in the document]
vaibhav-yb: Added the relavant distinction in the methods. `null` values will now be represented by a string "null" and empty values with empty string.

PR #274, opened 2023-10-05
Added retry mechanism for setCheckpoint in snapshot

Sumukh-Phalgaonkar:
## Problem
The setCheckpoint call in snapshot did not have proper retry mechanism in case of failure. As a result, in case of multi node connection when one of the node is down, the `setCheckpoint` call would fail.

## Solution
Added the requirred retry mechanism for `setCheckpoint` call.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -231,58 +231,7 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
      * @param tabletId
      * @throws Exception when the SetCheckpoint RPC fails on the service
      */
-    protected void setCheckpointWithRetryBeforeSnapshot(
-      String tableId, String tabletId) throws Exception {
-      short retryCount = 0;
-
-      // The SetCDCCheckPoint RPC, relies on a cache to obtain a list of all the tservers.
-      // In case of multi host port connection url, if one of the DB node goes down,
-      // it takes some time for the cache to refresh and return correct tserver list.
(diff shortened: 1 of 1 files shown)


suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -341,7 +353,7 @@ protected boolean isSnapshotRequired(GetCheckpointResponse getCheckpointResponse
         // A call to set the checkpoint is required first otherwise we will get an error 
         // from the server side saying:
         // INTERNAL_ERROR[code 21]: Stream ID {} is expired for Tablet ID {}
-        setCheckpointWithRetryBeforeSnapshot(tableId, tabletId);
+        setCheckpointWithRetry(tableId, tabletId, 0, 0, true, false, 0L);
```
vaibhav-yb, line 356: 1. Let's do one thing, rename this method to `markStreamActive`

2. I think a more preferable way to do that would be to use just the boolean condition, for example, `bootstrap` and remove the redundant parameters which can be inferred from the flag itself:

```java
public void markStreamActive(String tableId, String tabletId, bootstrap)  {
  try {
    long term = 0;
    long index = 0;
    if (bootstrap) {
      term = -1;
      index = -1;
    }
 
    YBClientUtils.setCheckpoint(this.syncClient, this.connectorConfig.streamId(), tableId, tabletId, term,
                                index, true /* initialCheckpoint */, bootstrap);
  } catch (Exception e) {
    // Exception handling
  }
}

```

vaibhav-yb (changes requested):

vaibhav-yb (approved):

suranjan (approved):
much cleaner

PR #277, opened 2023-10-06
[yugabyte/yugabyte-db#12751] Changes to add configurable rpc retry attempts

Sumukh-Phalgaonkar:
Currently, the AsynClient retries an RPC for 100 times within an interval of almost one minute. This results in `Too many attempts` exception much before the specified time out is reached. 

This PR contains changes to add configurable maximum number of rpc retry attempts, while creating `AsyncClient`. The default number of retries have been set to a sufficiently large number, such that all the attempts will not get over much before the specified timeout.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
         <version.kafka>3.3.1</version.kafka>
         <version.org.slf4j>1.7.36</version.org.slf4j>
         <version.logback>1.4.0</version.logback>
-        <version.ybclient>0.8.67-SNAPSHOT</version.ybclient>
+        <version.ybclient>0.8.67-20231010.165133-1</version.ybclient>
         <version.gson>2.8.9</version.gson>
 
         <!--
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
         <version.kafka>3.3.1</version.kafka>
         <version.org.slf4j>1.7.36</version.org.slf4j>
         <version.logback>1.4.0</version.logback>
-        <version.ybclient>0.8.64-20230929.065833-2</version.ybclient>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -47,7 +47,7 @@
         <version.kafka>3.3.1</version.kafka>
         <version.org.slf4j>1.7.36</version.org.slf4j>
         <version.logback>1.4.0</version.logback>
-        <version.ybclient>0.8.64-20230929.065833-2</version.ybclient>
+        <version.ybclient>0.8.65-SNAPSHOT</version.ybclient>
```
vaibhav-yb, line 50: YBClient service is already on `0.8.66` - make sure you increase your version accordingly.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnector.java
```diff
@@ -201,7 +201,7 @@ private void closeYBClient() {
     }
 
     private YBClient getYBClientBase(String hostAddress, long adminTimeout, long operationTimeout, long socketReadTimeout,
-                                     int maxNumTablets, String certFile, String clientCert, String clientKey) {
+                                     int maxNumTablets, String certFile, String clientCert, String clientKey, int maxAttempts) {
```
vaibhav-yb, line 204: If I remember correctly, this method is just used in `YugabyteDBConnector` - we can remove this method now and use the one in `YBClientUtils` as the later one only takes connector configuration as a parameter. We can then use the later method everywhere.

vaibhav-yb (changes requested):

suranjan (approved):

vaibhav-yb (approved):

PR #289, opened 2023-10-31
[yugabyte/yugabyte-db#19260] Add support for NOTHING and DEFAULT on connector

thedhruvrawat:
Adding support for two new record types: `DEFAULT` and `NOTHING` based on Postgres replica identity types. Their behaviour for the basic queries listed below is as follows:

**Queries**:
```
CREATE TABLE t1(id INT PRIMARY KEY, v1 INT, v2 INT);
INSERT INTO t1 VALUES (1,2,3);
# Perform an update
UPDATE t1 SET v1 = 3 WHERE id = 1;
# Perform a delete
DELETE from t1 WHERE id = 1;
```

| `RECORD_TYPE` | OPERATION | OLD_IMAGE | NEW_IMAGE  |
| ----------------- | ------------ | ------------ | -------------  | 
| `DEFAULT`           | UPDATE        | `{{}}`            | `{{1},{3},{3}}` |
|                               | DELETE        | `{{1}}`           | `{{}}`              |
| `NOTHING`          | UPDATE        | `{{}}`            | `{{1},{3},{3}}` |
|                               | DELETE        | `{{}}`            | `{{}}`               |

Also, the `ALL` and `MODIFIED_COLUMNS_OLD_AND_NEW_IMAGES` modes have been renamed to `FULL` and `CHANGE_OLD_NEW` respectively.

Have kept the old `ALL` and `MODIFIED_COLUMNS_OLD_AND_NEW_IMAGES` to maintain backward compatibility.

### Test Plan
Added tests for `NOTHING` and `DEFAULT` modes
- `mvn clean test -Dtest=YugabyteDBBeforeImageTest#consecutiveSingleShardTransactionsForDefault`
- `mvn clean test -Dtest=YugabyteDBBeforeImageTest#consecutiveSingleShardTransactionsForNothing`

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/HelperBeforeImageModes.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/HelperBeforeImageModes.java
@@ -4,53 +4,53 @@ import io.debezium.config.EnumeratedValue;
 public class HelperBeforeImageModes {
 
     public enum BeforeImageMode implements EnumeratedValue {
-      /**
-       * [Old type that is no longer allowed to be created] ALL mode, both old and new images of the
-       * item
-       */
-      ALL("ALL"),
+        /**
+         * [Old type that is no longer allowed to be created] ALL mode, both old and new images of the
+         * item
+         */
+        ALL("ALL"),
 
-      /**
-       * CHANGE mode (default), only the changed columns
-       */
-      CHANGE("CHANGE"),
+        /**
+         * CHANGE mode (default), only the changed columns
+         */
+        CHANGE("CHANGE"),
 
-      /**
-       * [Old type that is no longer allowed to be created] FULL_ROW_NEW_IMAGE mode, the entire
-       * updated row as new image, entire row as old image for DELETE
-       */
-      FULL_ROW_NEW_IMAGE("FULL_ROW_NEW_IMAGE"),
+        /**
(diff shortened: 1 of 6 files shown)


src/main/java/io/debezium/connector/yugabytedb/HelperBeforeImageModes.java
```diff
@@ -21,9 +21,19 @@ public enum BeforeImageMode implements EnumeratedValue {
         FULL_ROW_NEW_IMAGE("FULL_ROW_NEW_IMAGE"),
 
         /**
-         * MODIFIED_COLUMNS_OLD_AND_NEW_IMAGES mode, old and new images of modified column
+         * CHANGE_OLD_NEW mode, old and new images of modified column
```
vaibhav-yb, line 24: Should we keep the `ALL` and `MODIFIED_COLUMNS_OLD_AND_NEW_IMAGES` to maintain backward compatibility?

dr0pdb: Yes, we should. I commented the same on the server side diff. It is required since we store this in sys-catalog on disk.

thedhruvrawat: Restored. Have added [Deprecated] in comment

vaibhav-yb (approved):

PR #291, opened 2023-10-31
[yugabyte/yugabyte-db#15603] Changes to use tablet ranges instead of tabletIDs

vaibhav-yb:
## Problem

The model today uses the top level `YugabyteDBConnector` class to divide the tablets into tasks so that they can be polled parallelly. The list of tablets is sent through a serialized string sent as a connector configuration property which is then deserialized by the tasks to retrieve the original list and stored as a copy eg. `tabletPairList`.

Now whenever there is a tablet split, we remove the split parent from `tabletPairList` and add the children tablets to it so that they can be polled further - note that we cannot alter the serialized string config available to the task as that is immutable. In case of connector restarts, the distribution of tablets is handled properly but if there is a task restart, the task goes and reads the config to deserialize the list again and poll on those tablets, but since the original tablet from this immutable list has been split and we now have children tablets, we will not get any relevant information for this split tablet from the service and the connector would fail.

## Solution

This PR introduces a change in the way the tablets are sent to the tasks. Instead of sending just tabletIDs, we also send the tablet ranges now. This model gives us the insurance that even after the parent is split, the children tablets we will receive will be in the original tablet range and our connector will keep functioning smoothly.

For example, suppose we have our tablet ranges in alphabetical order `A, B, C` and we have a parent tablet `[A, C)` so at the task level, we will send the config with this range and if this gets split into two children i.e. `[A, B)` and `[B, C)`, they will still belong to the original range.

#### HashPartition

For facilitation of the tablets, we have introduced a new class `HashPartition` which will be used to represent a tablet on the service side along with its partition key ranges i.e. start and end.

There are some helper methods included to help us evaluate whether a tablet is conflicting with another or whether some tablet is parent for another i.e. `HashPartition#containsPartition`

### Test plan

Added a test to validate the methods in the class `HashPartition`, to run:

```sh
mvn clean test -Dtest=HashPartitionTest
```

### Relevant GitHub issues
* https://github.com/yugabyte/yugabyte-db/issues/19698
* https://github.com/yugabyte/yugabyte-db/issues/19833
* https://github.com/yugabyte/yugabyte-db/issues/15603
* https://github.com/yugabyte/yugabyte-db/issues/19411
* https://github.com/yugabyte/yugabyte-db/issues/19687


src/main/java/io/debezium/connector/yugabytedb/connection/HashPartition.java
```diff
@@ -0,0 +1,390 @@
...
+	 * Compare two partitions
+	 * @param other {@link HashPartition} to compare with
+	 * @return true if other partition is contained in current, false otherwise
+	 */
+	public boolean containsPartition(HashPartition other) {
+		if (!this.tableId.equals(other.tableId)) {
+			return false;
+		}
+
+		/* We know that this partition can contain another for sure in case of following:
```
suranjan, line 184: Have we written this logic or borrowed it from somewhere else?

src/main/java/io/debezium/connector/yugabytedb/connection/HashPartition.java
```diff
@@ -0,0 +1,390 @@
...
+						 && (compareKey(this.partitionKeyEnd, other.partitionKeyStart) > 0)
+						 && (compareKey(this.partitionKeyEnd, other.partitionKeyEnd) > 0);
+	}
+
+	/**
+	 *
+	 * @param other the {@link HashPartition} to compare with
+	 * @return true if the partitions have conflicting boundaries, false otherwise
+	 */
+	public boolean isConflictingWith(HashPartition other) {
```
suranjan, line 225: Is this logic borrowed or we have written it ourselves?

src/main/java/io/debezium/connector/yugabytedb/connection/HashPartition.java
```diff
@@ -0,0 +1,390 @@
...
+		}
+
+		return result;
+	}
+
+	/**
+	 * Validate that the partitions we have make up the full exhaustive range of keys.
+	 * @param hashPartitions a list of all the HashPartitions
+	 */
+	public static void validateCompleteRanges(List<HashPartition> hashPartitions) {
```
suranjan, line 327: Is this logic borrowed or our own?

src/main/java/io/debezium/connector/yugabytedb/connection/HashPartition.java
```diff
@@ -0,0 +1,390 @@
...
+ * Partition class to store a representation of a tablet with its ranges. Most of the logic for
+ * this class has been taken from
+ * <a href="https://github.com/yugabyte/yugabyte-db/blob/master/java/yb-client/src/main/java/org/yb/client/Bytes.java">Bytes.java</a><br><br>
+ *
+ * Part of the logic to compare the two partitions has been borrowed from
+ * <a href="https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/util/slice.h#L311">slice.h</a>
+ * @author Vaibhav Kushwaha (vkushwaha@yugabyte.com)
+ */
+public class HashPartition implements Comparable<HashPartition> {
+	private final String tableId;
```
suranjan, line 27: Some issue with formatting of this file

src/main/java/io/debezium/connector/yugabytedb/connection/YBTablet.java
```diff
@@ -0,0 +1,45 @@
...
+		return partitionKeyStart;
+	}
+
+	public byte[] getPartitionKeyEnd() {
+		return partitionKeyEnd;
+	}
+
+	public HashPartition toHashPartition() {
+		return new HashPartition(tableID, tabletID, partitionKeyStart, partitionKeyEnd,
+														 new ArrayList<>());
```
suranjan, line 43: add equality and hash functions as well

suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/connection/HashPartition.java
suranjan: is the formatting correct?

vaibhav-yb: The formatting seems to be fine in my IDE, probably something off with the GH preview. Will do a double check once again.

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
```diff
@@ -185,6 +186,25 @@ public void testRecordConsumption() throws Exception {
...
+        Configuration.Builder configBuilder = TestHelper.getConfigBuilder("public.t1_range", dbStreamId);
+        startEngine(configBuilder);
+        final int recordsCount = 100;
+
+        awaitUntilConnectorIsReady();
+
+        TestHelper.executeBulk("INSERT INTO t1_range values (%d);", recordsCount);
+
+        waitAndFailIfCannotConsume(new ArrayList<>(), recordsCount);
+    }
```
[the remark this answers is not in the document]
vaibhav-yb: Yes, this one is following the indentation of the rest of the file. Fixed the issue of incorrect indentation in other files.

src/main/java/io/debezium/connector/yugabytedb/YBPartition.java
```diff
@@ -116,19 +116,19 @@ static class Provider implements Partition.Provider<YBPartition> {
...
                 throw new DebeziumException("Error while deserializing tablet list", e);
             }
 
             Set<YBPartition> partitions = new HashSet<>();
-            for (Pair<String, String> tabletPair : tabletPairList) {
-                partitions.add(new YBPartition(tabletPair.getLeft(), tabletPair.getRight()));
+            for (Pair<Pair<String, String>, Pair<String, String>> tabletPair : tabletPairList) {
+                partitions.add(new YBPartition(tabletPair.getLeft().getKey(), tabletPair.getLeft().getValue()));
             }
             LOGGER.debug("The partition being returned is " + partitions);
```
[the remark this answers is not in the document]
vaibhav-yb: This is a one time log but it is not useful to us at this stage as we are printing the same information at the top level connector and at the task levels as well.

src/main/java/io/debezium/connector/yugabytedb/connection/HashPartition.java
```diff
@@ -0,0 +1,400 @@
...
+      }
+    }
+
+    throw new DebeziumException("Given start key " + Arrays.toString(startKey)
+                                  + " not found in the list of partitions");
+  }
+
+  private static byte[] getByteArray(String str) {
+    // Length 2 also means it is an empty byte array as it would just contain "[]"
+    if (str.isEmpty() || str.length() == 2) {
```
suranjan, line 386: why not use equals with '[]'

suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -277,6 +281,18 @@ protected void markNoSnapshotNeeded(YBClient syncClient, YBTable ybTable, String
         }
     }
 
+    protected void getTabletPairListFromRange(GetTabletListToPollForCDCResponse r, List<Pair<String, String>> res) {
```
suranjan, line 284: rename to getAlltheHashPartForTask

suranjan (approved):

PR #295, opened 2023-11-09
[yugabyte/yugabyte-db#19294] Resolved the issue of large input strings getting truncated

Sumukh-Phalgaonkar:
## Problem
There was a bug in the `convertString()` method of `YugabyteDBCQLValueConverter` because of which large input strings were getting truncated to the first 50 characters. This was due to the wrong `toString()` method being used for conversion of ByteString to String.

## Solution
This issue was resolved by using `toStringUtf8()` instead of `toString()` method to extract the string from the ByteString. This PR also adds a test to verify the correct behaviour.

### DIFF
--- a/src/test/java/io/debezium/connector/yugabytedb/YugabyteDBCQLTest.java
+++ b/src/test/java/io/debezium/connector/yugabytedb/YugabyteDBCQLTest.java
@@ -292,8 +292,8 @@ public class YugabyteDBCQLTest extends YugabyteDBContainerTestBase {
     private String getBigInputString(int length) {
         StringBuilder sb = new StringBuilder();
         Random r = new Random();
-        for(int i = 0; i<length; i++) {
-            sb.append((char)(r.nextInt(26)+ 'A'));
+        for (int i = 0; i < length; i++) {
+            sb.append((char) (r.nextInt(26) + 'A'));
         }
         return sb.toString();
     }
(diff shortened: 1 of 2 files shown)


src/test/java/io/debezium/connector/yugabytedb/YugabyteDBCQLTest.java
```diff
@@ -266,4 +289,23 @@ private void getRecords(List<SourceRecord> records, long totalRecordsToConsume,
         assertEquals(totalRecordsToConsume, totalConsumedRecords.get());
     }
 
+    private String getBigInputString(int length) {
+        StringBuilder sb = new StringBuilder();
+        Random r = new Random();
+        for(int i = 0; i<length; i++) {
```
vaibhav-yb, line 295: correct the formatting here:
```
for (int i = 0; i < length; i++) {
  // loop content
}
```

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBCQLTest.java
```diff
@@ -266,4 +289,23 @@ private void getRecords(List<SourceRecord> records, long totalRecordsToConsume,
...
+            sb.append((char)(r.nextInt(26)+ 'A'));
```
vaibhav-yb, line 296: Correct formatting here too:
```
sb.append((char) (r.nextInt(26) + 'A'));
```

vaibhav-yb (approved):

PR #298, opened 2023-11-24
[yugabyte/yugabyyte-db#19811] Changes to support Publication and Replication slot

Sumukh-Phalgaonkar:
This PR contains changes needed to support Publication and Replication slot in the connector. The PR contains three major areas of changes. 

1.  Extracting the stream id from the Replication slot.
2. Extracting the table include list from the Publication.
3. Dynamically adding tables for streaming in case of `ALTER PUBLICATION`.

For all these changes we execute queries on the `pg_replication_slots` and `pg_publication_tables` views.


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnector.java
```diff
@@ -72,9 +73,23 @@ public void start(Map<String, String> props) {
         this.props = props;
         LOGGER.debug("Props " + props);
         Configuration config = Configuration.from(this.props);
+
+        String streamId = config.getString(YugabyteDBConnectorConfig.STREAM_ID);
+        String tableIncludeList = config.getString(YugabyteDBConnectorConfig.TABLE_INCLUDE_LIST);
+        
+        // For CQL tables as well as for the config not using Publication/Replication streamId will be non null
+        usePublication = streamId == null || streamId.isEmpty();
```
vaibhav-yb, line 81: I would prefer that `usePublication` be a configuration property rather than us evaluating based on this condition. It would have 2 direct impact:
1. The user will be knowing for sure that if they want to use publication then they can provide the config and skip stream ID. We can also add validation in the configuration class that is this `use.publication` or whatever name is given, if it is set to `true` then stream ID shouldn't be set and we can throw out the error early. I am saying this to keep it till we completely transition to the slot model.
2. Once that is done, we can simply use a getter method i.e. `config.shouldUsePublication()` everywhere we pass the variable `usePublication`

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnector.java
```diff
@@ -118,6 +133,16 @@ public List<Map<String, String>> taskConfigs(int maxTasks) {
                 } catch (IOException e) {
                     e.printStackTrace();
                 }
+
+                if(usePublication) {
```
vaibhav-yb, line 137: Formatting after if --> `if (usePublication) {`

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -356,6 +365,22 @@ public static SecureConnectionMode parse(String value, String defaultValue) {
...
+            }
+
+            @Override
+            public String getPostgresPluginName() {
+                return getValue();
+            }
+
+            @Override
+            public boolean supportsTruncate() {
+                return true;
```
vaibhav-yb, line 381: We don't recommend truncating the table for us, maybe set this to `false`

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -700,13 +725,45 @@ public JdbcConfiguration getJdbcConfig() {
...
+            .withDefault(ReplicationConnection.Builder.DEFAULT_SLOT_NAME)
+            .withValidation(YugabyteDBConnectorConfig::validateReplicationSlotName)
+            .withDescription("The name of the YSQL logical decoding slot created for streaming changes from a plugin." +
+                    "Defaults to 'debezium");
+    
+    public static final Field DROP_SLOT_ON_STOP = Field.create("slot.drop.on.stop")
+            .withDisplayName("Drop slot on stop")
+            .withType(Type.BOOLEAN)
+            .withGroup(Field.createGroupEntry(Field.Group.CONNECTION_ADVANCED_REPLICATION, 3))
+            .withDefault(false)
```
vaibhav-yb, line 752: Create a final variable for the default values at the top and then use it here. That will ensure if we ever change the default value then it gets changed everywhere.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1249,6 +1307,26 @@ protected LogicalDecoder plugin() {
...
 
+    protected String slotName() {
+        return getConfig().getString(SLOT_NAME);
+    }
+
+    protected boolean dropSlotOnStop() {
+        if (getConfig().hasKey(DROP_SLOT_ON_STOP.name())) {
+            return getConfig().getBoolean(DROP_SLOT_ON_STOP);
+        }
+        // Return default value
```
vaibhav-yb, line 1318: Add an extra line before this comment.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1529,161 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+                        final Metronome retryMetronome = Metronome.parker(Duration.ofMillis(config.getLong(RETRY_DELAY_MS)), Clock.SYSTEM);
+                        retryMetronome.pause();
+                    } catch (InterruptedException ie) {
+                        LOGGER.warn("Connector retry sleep interrupted by exception: {}", ie);
+                        Thread.currentThread().interrupt();
+                    }
+                }
+            }
+        }
+        if (exception != null) {
```
vaibhav-yb, line 1572: Add extra line before and after the block.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1529,161 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+            while (retryCount <= config.getInteger(MAX_RETRIES)) {
+                try (Connection connection = ybConnection.connection();
+                      final Statement statement = connection.createStatement()) {
+                    String query = "SELECT yb_stream_id FROM pg_replication_slots WHERE slot_name = '"+ config.getString(SLOT_NAME) + "';";
+                    final ResultSet rs = statement.executeQuery(query);
+                    if (rs.next()) {
+                        String streamId = rs.getString("yb_stream_id");
+                        return streamId;
+                    } else {
+                        LOGGER.error("StreamId not found");
```
vaibhav-yb, line 1546: Add more detail in the error log i.e. Stream ID not found in the replication slots.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1529,161 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+                } catch (InterruptedException ie) {
+                    LOGGER.warn("Connector retry sleep interrupted by exception: {}", ie);
+                    Thread.currentThread().interrupt();
+                }
+            } catch (DebeziumException e) {
+                exception = e;
+                throw e;
+            }
+        }
+        if(exception != null) {
```
vaibhav-yb, line 1645: Add extra line before and after block. Also formatting around `if`

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBTablePoller.java
```diff
@@ -87,69 +92,142 @@ public void run() {
...
-      if (cachedTableInfoSet == null) {
-        LOGGER.debug("Cached table list in the poller thread is null, initializing it now");
-        cachedTableInfoSet = resp.getTableInfoList().stream().collect(Collectors.toSet());
-      } else {
-        if (cachedTableInfoSet.size() != resp.getTableInfoList().size()) {
-          Set<TableInfo> tableInfoSetFromResponse =
-            resp.getTableInfoList().stream().collect(Collectors.toSet());
+        return shouldRestart;
+      } catch (Exception e) {
+        ++retryCount;
```
vaibhav-yb, line 147: I do not understand the above change, might be because I am missing some info here. Can you please explain a little?

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBPublicationReplicationTest.java
```diff
@@ -0,0 +1,193 @@
...
+import io.debezium.transforms.ExtractNewRecordStateConfigDefinition;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class YugabyteDBPublicationReplicationTest extends YugabytedTestBase {
+
+    public static String createPublicationForTableStatement = "CREATE PUBLICATION %s FOR TABLE %s ;";
+    public static String createPublicationForALLTablesStatement = "CREATE PUBLICATION %s FOR ALL TABLES ;";
+    public static String createReplicationSlotStatement = "SELECT pg_create_logical_replication_slot('test_replication_slot', 'yboutput');";
+    public static String dropReplicationSlotStatement = "SELECT pg_drop_replication_slot('test_replication_slot');";
```
vaibhav-yb, line 37: If the above statements are generic and will be useful in other classes as well, we can put them to `TestHelper` as well.

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBPublicationReplicationTest.java
```diff
@@ -0,0 +1,193 @@
...
+import org.yb.client.GetDBStreamInfoResponse;
+import org.yb.client.YBClient;
+
+import io.debezium.config.Configuration;
+import io.debezium.connector.yugabytedb.transforms.YBExtractNewRecordState;
+import io.debezium.transforms.ExtractNewRecordStateConfigDefinition;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class YugabyteDBPublicationReplicationTest extends YugabytedTestBase {
```
vaibhav-yb, line 32: The class to be extended would be `YugabyteDBContainerTestBase`

Also, put a javadoc comment telling what this test is for and what is the minimum service version required for it to run.

vaibhav-yb (changes requested):
Have we checked that these changes work with the existing model as well? Is backward compatibility ensured?

Also, I remember you were talking about borrowing some code from Postgres connector, can we put a comment citing the source?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
dr0pdb: This will happen if the replication slot with the given name doesn't exist. That is the only case, when there will be an empty result set.

So let's log the error that `Could not find the replication slot with name '__name__'`

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnector.java
Sumukh-Phalgaonkar: We wanted to make the connector configs to be exactly similar to the debezium postgres connector configs. Adding a new config property `usePublication` will create a difference between the two.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBTablePoller.java
Sumukh-Phalgaonkar: In the current implementation, in `areThereNewTablesInStream()` method, we initialize the retry count to zero at entry. Suppose we encounter any error inside the `try` block, we would move to the catch block, increment the retry count and return `false`.  When the method is again invoked after the polling interval, the retry count would be again set to zero. Hence the retries for the failure that was encountered initially, never took place.

In case of publications, we are polling the `pg_publication_tables` to see if new tables were added to the publication, instead of polling over `stream_info`

Sumukh-Phalgaonkar:
> Have we checked that these changes work with the existing model as well? Is backward compatibility ensured?
> 
> Also, I remember you were talking about borrowing some code from Postgres connector, can we put a comment citing the source?

Yes, the changes work with existing model as well. All the existing unit tests are working fine. We have only copied the config declaration from Postgres connector. The `YugabyteDBConnectorConfig` file already cites the copyright

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1395,6 +1482,20 @@ public static ConfigDef configDef() {
...
+        if (name != null) {
+            if (!name.matches("[a-z0-9_]{1,63}")) {
+                problems.accept(field, name,
+                        "Valid replication slot name must contain only digits, lowercase characters and underscores with length <= 63");
+                ++errors;
+            }
+        }
+        return errors;
+    }
+
```
dr0pdb, line 1498: Was this taken from upstream connector or added by you?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1534,165 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+                    .with(JdbcConfiguration.PASSWORD, config.getString(PASSWORD))
+                    .withDefault(YugabyteDBConnectorConfig.MAX_RETRIES, config.getInteger(MAX_RETRIES))
+                    .with(YugabyteDBConnectorConfig.RETRY_DELAY_MS, config.getLong(RETRY_DELAY_MS))
+                    .build();
+    }
+
+    public static boolean checkIfPublicationExists(String publicationName, YugabyteDBConnection ybConnection) throws SQLException {
+        try (Connection connection = ybConnection.connection();
+              final Statement statement = connection.createStatement()) {
+            String query = "SELECT * FROM pg_publication_tables WHERE pubname = '" + publicationName + "';";
```
dr0pdb, line 1675: `pg_publication_tables` lists all the tables that are part of a publication. There can be quite a few of them and we don't need to know them here.

Use `pg_publication` table here.
`SELECT * FROM pg_publication WHERE pubname = '__name__';`

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1534,165 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+                    }
+                    return fetchTableIncludeList(ybConnection, publicationName);
+                } else if (autoCreateMode.equalsIgnoreCase("filtered")) {
+                    String tableList = config.getString(TABLE_INCLUDE_LIST);
+                    if (tableList == null || tableList.isEmpty()) {
+                        throw new DebeziumException("No tables included for Filtered Publication");
+                    }
+
+                    String createQuery = "CREATE PUBLICATION " + config.getString(PUBLICATION_NAME) + " FOR TABLE " + tableList + ";";
+                    String alterQuery = "ALTER PUBLICATION " + config.getString(PUBLICATION_NAME) + " SET TABLE " + tableList + ";";
```
dr0pdb, line 1618: Was this "Alter publication" part also taken from the upstream connector?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1534,165 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+
+                try {
+                    final Metronome retryMetronome = Metronome.parker(Duration.ofMillis(config.getLong(CONNECTOR_RETRY_DELAY_MS)), Clock.SYSTEM);
+                    retryMetronome.pause();
+                } catch (InterruptedException ie) {
+                    LOGGER.warn("Connector retry sleep interrupted by exception: {}", ie);
+                    Thread.currentThread().interrupt();
+                }
+            } catch (DebeziumException e) {
+                exception = e;
```
dr0pdb, line 1648: This catch statement is redundant here. Please remove.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBTablePoller.java
```diff
@@ -87,69 +92,142 @@ public void run() {
    * @return true if there is a new table in the stream info
    */
   private boolean areThereNewTablesInStream() {
-    try {
-      boolean shouldRestart = false;
-      GetDBStreamInfoResponse resp = this.ybClient.getDBStreamInfo(this.connectorConfig.streamId());
+    // Reset the retry counter.
```
dr0pdb, line 95: Comment doesn't add much value. You can add more information about why the reset is important here or simply remove the comment.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBTablePoller.java
```diff
@@ -87,69 +92,142 @@ public void run() {
...
-            LOGGER.debug("New tables as received in the GetDBStreamInfoResponse: ");
-            difference.forEach(tableInfo -> {
-              LOGGER.debug(tableInfo.getTableId().toStringUtf8());
-            });
-          }
+    return false;
+  }
+
+  private boolean areThereNewTablesInPublication() {
+    // Reset the retry counter.
```
dr0pdb, line 165: same as above.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBTablePoller.java
```diff
@@ -87,69 +92,142 @@ public void run() {
...
+
+            Set<String> difference = new HashSet<>(responseSet);
+            difference.removeAll(cachedSet);
+
+            if (LOGGER.isDebugEnabled()) {
+              LOGGER.debug("Common tables between the cached table info set and the set received "
+                          + "from GetDBStreamInfoResponse:");
+              intersection.forEach(table -> LOGGER.debug(table));
+
+              LOGGER.debug("New tables as received in the GetDBStreamInfoResponse: ");
```
dr0pdb, line 193: Wrong log messages here and above. We are not using `GetDBStreamInfoResponse` here.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBTablePoller.java
```diff
@@ -181,6 +259,21 @@ public boolean isTableIncludedForStreaming(String tableUUID) throws Exception {
...
 
+  private Set<String> getTablesInPublication() throws Exception {
+    try (YugabyteDBConnection ybConnection = new YugabyteDBConnection(this.connectorConfig.getJdbcConfig(), YugabyteDBConnection.CONNECTION_GENERAL);
+        Connection connection = ybConnection.connection()) {
+          Set<String> tablesInPublication = new HashSet<String>();
+          Statement statement = connection.createStatement();
+          String getTablesFromPublicationQuery = "SELECT * FROM pg_publication_tables WHERE pubname = '"+this.connectorConfig.publicationName()+"' ;";
+          ResultSet rs = statement.executeQuery(getTablesFromPublicationQuery);
+          while(rs.next()) {
+            String tableName = rs.getString("tablename");
```
dr0pdb, line 270: We need to also keep the schema name like we do in the table include list.

src/main/java/io/debezium/connector/yugabytedb/connection/ReplicationConnection.java
```diff
@@ -103,7 +103,7 @@ interface Builder {
          */
         String DEFAULT_SLOT_NAME = "debezium";
         String DEFAULT_PUBLICATION_NAME = "dbz_publication";
-        boolean DEFAULT_DROP_SLOT_ON_CLOSE = true;
+        boolean DEFAULT_DROP_SLOT_ON_CLOSE = false;
```
dr0pdb, line 106: Is there a reason to change the default here?

dr0pdb (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
Sumukh-Phalgaonkar: This was taken from the upstream connector as a part of declaration of `slot.name` field

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
Sumukh-Phalgaonkar: We are throwing `DebeziumException` in certain cases like when autocreate is disabled and publication does not exist. We don't want to retry in these cases. That is why I had added a separate catch block for this.

dr0pdb: Yeah, but you are not doing anything here.

```
catch (some_exception) {
  // an assignment
  throw some_exception
}
```

The above part doesn't have any effect.

dr0pdb: Catching and rethrowing an exception = Not catching it

src/main/java/io/debezium/connector/yugabytedb/connection/ReplicationConnection.java
Sumukh-Phalgaonkar: The postgres connector has the default value of `DROP_SLOT_ON_CLOSE` as false, hence I changed it to false here.

Sumukh-Phalgaonkar: Checked in the upstream connector code, this field is set to true. While declaring the config property `drop.slot.on.stop` they are setting its default to false. Will make similar changes here.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
Sumukh-Phalgaonkar: No, what I meant is I haven't borrowed this code from the Debezium PG connector. Instead I have implemented the behaviour as mentioned in the documentation [here](https://debezium.io/documentation/reference/1.9/connectors/postgresql.html#postgresql-publication-autocreate-mode)

suranjan: @Sumukh-Phalgaonkar We should check what the PG connector does for this. Its better to borrow their code/approach as that would be time tested.

Sumukh-Phalgaonkar: Yes, we are using the similar approach as PG connector. They also run `ALTER PUBLICATION`

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -738,7 +799,7 @@ public enum AutoCreateMode implements EnumeratedValue {
         /**
          * Enable publication for all tables.
          */
-        ALL_TABLES("all_tables"),
+        ALL_TABLES("all tables"),
```
dr0pdb, line 802: Why are we changing this?

It is `all_tables` in the upstream connector: https://github.com/debezium/debezium/blob/616bb52f208dd5e0aad8d95b4091bedf2fd5ef51/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java#L556

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1535,209 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+        while (retryCount <= config.getInteger(MAX_CONNECTOR_RETRIES)) {
+            try (YugabyteDBConnection ybConnection = new YugabyteDBConnection(getJdbcConfig(config), YugabyteDBConnection.CONNECTION_GENERAL)) {
+                if (autoCreateMode.equalsIgnoreCase("disabled")) {
+                    if (!checkIfPublicationExists(publicationName, ybConnection)) {
+                        throw new DebeziumException(
+                         "Publication autocreation is disabled and the provided publication does not exist. Create the publication and restart the connector or change the publication.autocreate.mode");
+                    } else {
+                        return fetchTableIncludeList(ybConnection, publicationName);
+                    }             
+                } else if (autoCreateMode.equalsIgnoreCase("all tables")) {
```
dr0pdb, line 1595: Is it possible to use the constant `ALL_TABLES` from the enum `AutoCreateMode`?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1535,209 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+        if (streamId == null || streamId.isEmpty()) {
+            return true;
+        }
+
+        int retryCount = 0;
+        while (retryCount <= config.getInteger(YugabyteDBConnectorConfig.MAX_RETRIES)) {
+            try (YugabyteDBConnection ybConnection = new YugabyteDBConnection(getJdbcConfig(config),YugabyteDBConnection.CONNECTION_GENERAL);
+                  Connection connection = ybConnection.connection();
+                  Statement statement = connection.createStatement()) {
+                    String query = "SELECT * FROM pg_replication_slots WHERE yb_stream_id='"+streamId+"';";
```
dr0pdb, line 1696: nit: Here and everywhere else in the PR. Space before and after `+`.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBTablePoller.java
```diff
@@ -181,6 +257,22 @@ public boolean isTableIncludedForStreaming(String tableUUID) throws Exception {
...
+    try (YugabyteDBConnection ybConnection = new YugabyteDBConnection(this.connectorConfig.getJdbcConfig(), YugabyteDBConnection.CONNECTION_GENERAL);
+        Connection connection = ybConnection.connection()) {
+          Set<String> tablesInPublication = new HashSet<String>();
+          Statement statement = connection.createStatement();
+          String getTablesFromPublicationQuery = "SELECT * FROM pg_publication_tables WHERE pubname = '" + this.connectorConfig.publicationName() + "' ;";
+          ResultSet rs = statement.executeQuery(getTablesFromPublicationQuery);
+          while(rs.next()) {
+            String tableName = rs.getString("tablename");
+            String schemaName = rs.getString("schemaname");
+            tablesInPublication.add(schemaName+"."+tableName);
```
dr0pdb, line 270: nit: space before and after `+`.

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBPublicationReplicationTest.java
```diff
@@ -0,0 +1,214 @@
...
+        TestHelper.execute(TestHelper.createReplicationSlotStatement);
+
+        Configuration.Builder configBuilder = TestHelper.getConfigBuilderWithPublication("yugabyte", "pub", "test_replication_slot");
+        configBuilder.with(YugabyteDBConnectorConfig.PUBLICATION_AUTOCREATE_MODE, "all tables");
+        startEngine(configBuilder);
+        final int recordsCount = 10;
+
+        awaitUntilConnectorIsReady();
+
+        Thread.sleep(30000);
```
dr0pdb, line 124: Why the sleep?

dr0pdb (commented):
Looks good from a correctness point. I'll rely on Vaibhav for the best practices and structure.

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBPublicationReplicationTest.java
Sumukh-Phalgaonkar: This is to ensure that we insert the records after bootstrap is completed. Adding a code comment for this.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1541,342 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+                    e);
+        }
+    }
+
+    private static Set<TableId> determineCapturedTables(Configuration config) throws Exception {
+        YugabyteDBConnection ybConnection = new YugabyteDBConnection(getJdbcConfig(config),YugabyteDBConnection.CONNECTION_GENERAL);
+        Set<TableId> allTableIds = ybConnection.readTableNames(config.getString(DATABASE_NAME), null, null, null);
+
+        Set<TableId> capturedTables = new HashSet<>();
+        RelationalTableFilters tableFilter = new RelationalTableFilters(config, new SystemTablesPredicate(), x -> x.schema() + "." + x.table());
```
dr0pdb, line 1791: We also need to skip the tables without primary key otherwise the publication creation will fail.

Please also add a test case where we have 2 tables - one with and another without primary key. It should autocreate mode as filtered and the publication should not exist.

dr0pdb (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1545,331 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+        String createOrUpdatePublicationStmt;
+        try {
+            Set<TableId> tablesToCapture = determineCapturedTables(config);
+            tableFilterString = tablesToCapture.stream().map(TableId::toDoubleQuotedString).collect(Collectors.joining(", "));
+            if (tableFilterString.isEmpty()) {
+                throw new DebeziumException(String.format("No table filters found for filtered publication %s", publicationName));
+            }
+            createOrUpdatePublicationStmt = isUpdate ? String.format("ALTER PUBLICATION %s SET TABLE %s;", publicationName, tableFilterString)
+                    : String.format("CREATE PUBLICATION %s FOR TABLE %s;", publicationName, tableFilterString);
+            LOGGER.info(isUpdate ? "Updating Publication with statement '{}'" : "Creating Publication with statement '{}'", createOrUpdatePublicationStmt);
```
vaibhav-yb, line 1769: See if this looks more readable:

```java
if (isUpdate) {
  createOrUpdatePublicationStmt = String.format("ALTER PUBLICATION %s SET TABLE %s;", publicationName, tableFilterString);
  LOGGER.info("Updating Publication with statement '{}'", createOrUpdatePublicationStmt);
} else {
  createOrUpdatePublicationStmt = String.format("CREATE PUBLICATION %s FOR TABLE %s;", publicationName, tableFilterString);
  LOGGER.info("Creating Publication with statement '{}'", createOrUpdatePublicationStmt);
}
```

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1545,331 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+            stmt.execute(createOrUpdatePublicationStmt);
```
vaibhav-yb, line 1770: Any error handling for this execute?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1545,331 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+        }
+        catch (Exception e) {
```
vaibhav-yb, line 1772: Move `catch` to previous line.

vaibhav-yb (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1545,365 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+        }
+
+        try (YBClient ybClient = YBClientUtils.getYbClient(config)) {
+            GetDBStreamInfoResponse getDBStreamInfoResponse = ybClient.getDBStreamInfo(streamId);
+            // Note that we need to pass one table ID to the ListCDCStreamsRequest RPC call.
+            Objects.requireNonNull(getDBStreamInfoResponse.getTableInfoList().get(0));
+            ListCDCStreamsResponse resp;
+            resp = ybClient.listCDCStreams(
+                            getDBStreamInfoResponse.getTableInfoList().get(0).getTableId().toStringUtf8(),
+                            getDBStreamInfoResponse.getNamespaceId(), null);
```
dr0pdb, line 1662: Why are we passing the table id in listCDCStreams? It will work but can be very confusing for readers. Ideally, we should just list out all CDC streams for the namespace.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnector.java
```diff
@@ -72,9 +73,26 @@ public void start(Map<String, String> props) {
...
+        String streamId = config.getString(YugabyteDBConnectorConfig.STREAM_ID);
+        String tableIncludeList = config.getString(YugabyteDBConnectorConfig.TABLE_INCLUDE_LIST);
+        
+        usePublication = YugabyteDBConnectorConfig.shouldUsePublication(config);
+
+        if (usePublication) {
+            YugabyteDBConnectorConfig.initPublication(config);
+            tableIncludeList = YugabyteDBConnectorConfig.extractTableListFromPublication(config);
+            YugabyteDBConnectorConfig.initReplicationSlot(config);
+            streamId =YugabyteDBConnectorConfig.extractStreamIdFromSlot(config);
```
dr0pdb, line 86: nit: space after `=`.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1433,6 +1545,365 @@ private static int validateLogicalDecodingMessageExcludeList(Configuration confi
...
+                final ResultSet rs = statement.executeQuery(query);
+                return rs.next();
+            } catch (Exception e) {
+                    retryCount++;
+                    if (retryCount > maxAttempts) {
+                        LOGGER.error("Failed to query pg_replication_slots. All {} attempts failed.", maxAttempts);
+                        throw new DebeziumException(e);
+                    }
+                    exception = new RuntimeException(e);
+                    LOGGER.warn("Error while querrying pg_replication_slots. Will retry, attempt {} out of {}", retryCount, maxAttempts);
```
dr0pdb, line 1885: typo: It should be "querying"

dr0pdb (approved):
Looks good. Minor comments.

vaibhav-yb (approved):

PR #302, opened 2023-12-05
Shift tablet removal from shouldWaitForCallback set logic to snapshot consumption phase

siddharth2411:
### Problem
Purpose of `shouldWaitForCallback` set:
If a tablet is part of this set, it means the connector should poll it for snapshot records and wait for kafka's callback. Therefore, empty tablets should not be a part of this set. 

Tablet removal Logic:
Currently, the logic to remove tablets from `shouldWaitForCallback` set is done after every getChanges call including the 1st Getchanges i.e. snapshot bootstrap call. We conclude there was no record seen in the GetChanges call based on the value of `lastRecordCheckpoint`. This variable is initialised based on the GetCheckpoint's response before we make any GetChanges call. 

GetCheckpoint returns opid > 0 --> set lastRecordCheckpoint to this valid Opid.
GetCheckpoint returns opid <= 0 --> set lastRecordCheckpoint to snapshotStartLsn i.e. opid = -1.-1 

- NoRecordSeen() checks if `lastRecordCheckpoint` is still equal to snapshotStartLsn. If NoRecordSeen() returns true, we remove the tablet from `shouldWaitForCallback` set.

- We update `lastRecordCheckpoint` only if we receive any READ records.

There are 2 issues with the current implementation of tablet remove logic:
1. Non-empty tablets getting removed
In the response of snapshot bootstrap call, we only receive a checkpoint and no records. At this point, since we didnt receive any records, the `lastRecordCheckpoint` is still set to `snapshotStartLsn` for all tablets (empty or non-empty). So, based on the current logic, we will remove all the tablets including non-empty tablets as well. This behavior is incorrect since we should only remove empty tablets. 

2. Empty tablets not getting removed
Consider this scenario: Connector has just finished snapshot bootstrap call on an empty tablet but havent yet sent the snapshot done key to the service. At this point, if the connector restarts, the `lastRecordCheckpoint` will be initialised with a value different than `snapshotStartLsn` since GetCheckpoint call on the empty tablet would returns opid > 0. So, in such cases, the noRecordSeen() would return false for empty tablets. Hence, we would not remove it from this set. 

### Solution

- Firstly, shift the tablet removal logic from the snapshot bootstrap call to the snapshot consumption phase i.e. 2nd GetChanges onwards since we receive records (if any) only during the snapshot consumption phase. Snapshot bootstrap call can be distinguished from Snapshot consumption call based on the from_op_id sent in the GetChanges request. During snapshot bootstrap, from_op_id is equal to `snapshotStartLsn` i.e term = -1, index = -1, snapshot_key = "", write_id = -1, snapshot_time = 0. But during snapshot consumption, term  > 0, index > 0, write_id = -1, snapshot_time > 0. Even empty tablets, after snapshot bootstrap is completed, have from_op_id > 0 because they do contain a DDL record.

- Secondly, we can distinguish between empty & non-empty tablets by simply keeping a count of READ records received during snapshot consumption phase. 

### Testing
Performed manual testing:

- Ran a test on single empty table split into 1 tablet. Restarted the connector after snapshot bootstrap and manually verified that tablet removal logic was executed during snapshot consumption phase and the empty tablet was removed.

- Ran a test on a single non-empty table split into 1 tablet. Manually verified that tablet removal logic was executed during snapshot consumption phase and no tablet was removed.

Note: Since this solution has changed the behaviour for tablet removal, some existing snapshot tests are failing because connector is not switching to streaming phase. This is primarily because tablets that are part of `shouldWaitForCallback` set are added to `tabletsWaitingForCallback` set. Tablet removal from `tabletsWaitingForCallback` set is also incorrect. This is a known [issue](https://github.com/yugabyte/yugabyte-db/issues/20135). 

### Linked Issues

- [[CDCSDK] Improve logic for adding tablets to snapshotCompletedTablets set](https://yugabyte.atlassian.net/browse/DB-9073)
- [[CDCSDK] After snapshot bootstrap, GetTabletListToPoll returns same tablet_id multiple times for colocated tables on connector restart](https://yugabyte.atlassian.net/browse/DB-9017)

### DIFF
--- a/src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotTest.java
+++ b/src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotTest.java
@@ -40,6 +40,8 @@ public class YugabyteDBSnapshotTest extends YugabyteDBContainerTestBase {
     public void before() throws Exception {
         initializeConnectorTestFramework();
         TestHelper.dropAllSchemas();
+        YugabyteDBSnapshotChangeEventSource.FAIL_AFTER_SETTING_INITIAL_CHECKPOINT = false;
+        YugabyteDBSnapshotChangeEventSource.FAIL_AFTER_BOOTSTRAP_GET_CHANGES = false;
     }
 
     @AfterEach
@@ -49,6 +51,8 @@ public class YugabyteDBSnapshotTest extends YugabyteDBContainerTestBase {
         TestHelper.executeDDL("drop_tables_and_databases.ddl");
         TestHelper.dropAllSchemas();
         resetCommitCallbackDelay();
+        YugabyteDBSnapshotChangeEventSource.FAIL_AFTER_SETTING_INITIAL_CHECKPOINT = false;
+        YugabyteDBSnapshotChangeEventSource.FAIL_AFTER_BOOTSTRAP_GET_CHANGES = false;
     }
 
     @AfterAll
--- a/src/main/java/io/debezium/connector/yugabytedb/SourceInfo.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/SourceInfo.java
@@ -135,16 +135,13 @@ public final class SourceInfo extends BaseSourceInfo {
      * Compares the lastRecordCheckpoint with {@code snapshotStartLsn} and {@code streamingStartLsn}.
      * If it is equal then it means that we haven't received any record for the given partition,
      * in case there's any difference, it technically siginifies that some record has updated the
-     * values. Also checks if commitTime is null. This covers cases when connector restarts
-     * after snapshot bootstrap is completed. In such cases, lastRecordCheckpoint for empty tablets
-     * would have been initialized with a non-null value.
+     * values.
      * @return true if the partition hasn't seen any record yet, false otherwise
      */
     public boolean noRecordSeen() {
         return (lastRecordCheckpoint == null)
             || lastRecordCheckpoint.equals(YugabyteDBOffsetContext.snapshotStartLsn())
-            || lastRecordCheckpoint.equals(YugabyteDBOffsetContext.streamingStartLsn())
-            || (commitTime == null);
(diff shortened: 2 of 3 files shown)


src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotTest.java
```diff
@@ -40,6 +40,8 @@ public static void beforeClass() throws Exception {
     public void before() throws Exception {
         initializeConnectorTestFramework();
         TestHelper.dropAllSchemas();
+        YugabyteDBSnapshotChangeEventSource.FAIL_AFTER_SETTING_INITIAL_CHECKPOINT = false;
```
suranjan, line 43: Don't see any specific test for this?

suranjan (changes requested):

vaibhav-yb (approved):
Approving with some minor changes requested.

siddharth2411:
> Some open questions:
> 
> 1. Currently, we add all the tablets to the `shouldWaitForCallback` set including the ones on which we the user has not asked us to take snapshot ([code](https://github.com/yugabyte/debezium-connector-yugabytedb/blob/0382188d7479f1f6d7812922250f50e50827346a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java#L455C9-L455C46)). So, Should we shift the limit of addition of tablets to this set to only tablets meant for snapshot?
> 2. Also, once we mark snapshot done on service for a tablet, should we remove the tablet's entry from this set if its present? (after this [line](https://github.com/yugabyte/debezium-connector-yugabytedb/blob/0382188d7479f1f6d7812922250f50e50827346a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java#L683))

This things have been added as part of this PR.

siddharth2411:
Changes in this PR were merged as part of https://github.com/yugabyte/debezium-connector-yugabytedb/pull/306. Closing this PR.

PR #303, opened 2023-12-07
[yugabyte/yugabyte-db#20177] Support to run the connector tests with consistent snapshot streams

Sumukh-Phalgaonkar:
Recently consistent snapshot streams have been introduced on the server side. All the connector tests currently run with the older streams. This PR provides support to run all the connector tests using older config model (`stream.id` and `table.include.list`) with the newly introduced consistent snapshot streams along with the older streams.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
         <version.kafka>3.3.1</version.kafka>
         <version.org.slf4j>1.7.36</version.org.slf4j>
         <version.logback>1.4.0</version.logback>
-        <version.ybclient>0.8.75-SNAPSHOT</version.ybclient>
+        <version.ybclient>0.8.75-20231215.051722-1</version.ybclient>
         <version.gson>2.8.9</version.gson>
 
         <!--
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
         <version.kafka>3.3.1</version.kafka>
         <version.org.slf4j>1.7.36</version.org.slf4j>
         <version.logback>1.4.0</version.logback>
-        <version.ybclient>0.8.73-20231201.070308-1</version.ybclient>
+        <version.ybclient>0.8.74-SNAPSHOT</version.ybclient>
         <version.gson>2.8.9</version.gson>
 
         <!--
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
         <version.kafka>3.3.1</version.kafka>
         <version.org.slf4j>1.7.36</version.org.slf4j>
         <version.logback>1.4.0</version.logback>
-        <version.ybclient>0.8.74-SNAPSHOT</version.ybclient>
+        <version.ybclient>0.8.75-SNAPSHOT</version.ybclient>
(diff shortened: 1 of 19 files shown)


pom.xml
```diff
@@ -47,7 +47,7 @@
         <version.kafka>3.3.1</version.kafka>
         <version.org.slf4j>1.7.36</version.org.slf4j>
         <version.logback>1.4.0</version.logback>
-        <version.ybclient>0.8.73-20231201.070308-1</version.ybclient>
+        <version.ybclient>0.8.74-SNAPSHOT</version.ybclient>
```
vaibhav-yb, line 50: Do take care of the version before merging.

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotTest.java
```diff
@@ -809,6 +820,68 @@ public void snapshotTwoColocatedNonEmptyAndNonColocatedEmptyThenStream(boolean c
         assertEquals(401, recordsForNonColocated.size());
     }
 
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    public void snapshotTwoColocatedNonEmptyAndNonColocatedEmptyThenStreamWithConsistentSnapshot(boolean colocation) throws Exception {
+        /* The objective of this test is to verify that we are able to consume all snapshot records on a
```
vaibhav-yb, line 826: Multi line comments follow this pattern:
```java
/*
 * Starting line of the comment would be here
 * and it will continue. You can end it on the next line as shown.
 */
```

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotTest.java
```diff
@@ -886,15 +959,75 @@ public void snapshotTwoColocatedNonEmptyAndNonColocatedNonEmptyThenStream(boolea
 
     @ParameterizedTest
     @ValueSource(booleans = {true, false})
-    public void verifyConnectorFailsIfMarkSnapshotDoneFails(boolean colocation) throws Exception {
+    public void snapshotTwoColocatedNonEmptyAndNonColocatedNonEmptyThenStreamWithConsistentSnapshot(boolean colocation) throws Exception {
+        /* The objective of this test is to verify that we are able to consume all snapshot records on a
```
vaibhav-yb, line 963: nit: fix comment style

vaibhav-yb: Moreover, this test only differs with the previous one in the sense that the non-colocated tables are non empty here. Is it possible to parameterise? If it makes things too complicated, leave it like this.

vaibhav-yb (changes requested):
I see that the tserver flag `TEST_yb_enable_cdc_consistent_snapshot_streams=true` is present in all the test. We can modify the method `initializeYBContainer` in `YugabyteDBContainerTestBase` to add this flag by default as we do for one master flag there.

vaibhav-yb (approved):

PR #304, opened 2023-12-08
Added identification marker for last snapshot record of the last batch

siddharth2411:
### Problem

There are 2 types of checkpoints received in the GetChanges response: Record's checkpoint and Response checkpoint. The service, while sending the last batch of snapshot records, sends the snapshot complete marker (key = "", write_id = 0, snapshot_time = 0) only in the response checkpoint. No record checkpoint will ever contain the snapshot complete marker.

Logic for removing tablet from `tabletsWaitingForCallback` set:
Tablets that are part of `tabletsWaitingForCallback` set are only removed when we receive snapshot complete marker (key = "", write_id = 0, snapshot_time = 0) in kafka's callback.  But the problem is, from the connector, we only send the record's checkpoint to kafka. So, as discussed above, we would never receive snapshot complete marker in kafka's callback. Therefore, we would never remove the tablets from this set and the connector would be stuck in the snapshot phase. We need a different identification marker for the last snapshot record of the last batch so that we can remove tablets on receiving kafka's callack for the last record.

### Solution
Modify the record checkpoint of last snapshot record of the last batch. The last record's checkoint will be set to the following:

- term = Long.MAX
- index = Long.MAX
- write_id = Integer.MAX
- time = Long.MAX
- snapshot_key = "LAST_SNAPSHOT_RECORD

Use the same field values to validate and remove tablets from the `tabletsWaitingForCallback` set accordingly. 

We also need to ensure that this record checkpoint is never sent as the explicit checkpoint in a GetChanges call. So, if we have received this modified checkpoint from kafka, we should stop polling on such tablets since kafka has already consumed all the snapshot records. 

NOTE: This solution works only in conjunction with this [PR](https://github.com/yugabyte/debezium-connector-yugabytedb/pull/302) because based on this PR's logic, we add tablets to the  `tabletsWaitingForCallback` set.

### Testing

- Ran the existing snapshot tests  
- Added two new tests where we restart connector after snapshot bootstrap and verify that we receive all records (snapshot + streaming).

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -508,7 +508,7 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
                   // If the checkpoint in tabletToExplicitCheckpoint map corresponds to the last record's checkpoint
                   // of last snapshot batch, it means kafka has consumed the last record. So, we should
                   // not poll on this tablet anymore, instead mark snapshot done for this tablet.
-                  if(checkpoint != null && IsLastSnapshotRecordOfLastBatch(OpId.from(checkpoint))) {
+                  if (checkpoint != null && isLastSnapshotRecordOfLastBatch(OpId.from(checkpoint))) {
                     LOGGER.debug("Already received the last record's modified checkpoint in kafka's callback. " +
                             "Discontinuing polling on tablet {}", part.getId());
                     LOGGER.info("Adding {} to the list of snapshot completed tablets", part.getId());
@@ -552,7 +552,7 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
 
                 // Process the response
                 for (int idx = 0; idx < resp.getResp().getCdcSdkProtoRecordsList().size(); idx++) {
-                  CdcService.CDCSDKProtoRecordPB record = resp.getResp().getCdcSdkProtoRecordsList().get(idx);
+                  CdcService.CDCSDKProtoRecordPB record = resp.getResp().getCdcSdkProtoRecords(idx);
                   CdcService.RowMessage m = record.getRowMessage();
                   YbProtoReplicationMessage message =
                     new YbProtoReplicationMessage(m, this.yugabyteDbTypeRegistry);
@@ -615,9 +615,9 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
                         }
                         Objects.requireNonNull(tId);
 
-                        // if the DML record is the last snapshot record of the last snapshot batch, set all the fields
+                        // If the DML record is the last snapshot record of the last snapshot batch, set all the fields
                         // of Opid to max values and snapshot key to LAST_SNAPSHOT_RECORD.
-                        if(isSnapshotCompleteMarker(finalOpId) &&
+                        if (isSnapshotCompleteMarker(finalOpId) &&
                                 (idx == (resp.getResp().getCdcSdkProtoRecordsList().size() - 1))) {
                           LOGGER.info("Modifying record checkpoint for last snapshot record of the last snapshot batch");
                           setIdentificationMarkerForLastSnapshotRecord(lsn);
@@ -788,7 +788,7 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
       }
 
       OpId opId = OpId.from(this.tabletToExplicitCheckpoint.get(partition.getId()));
-      if (IsLastSnapshotRecordOfLastBatch(opId)) {
+      if (isLastSnapshotRecordOfLastBatch(opId)) {
         LOGGER.info("Adding tablet {} to snapshot completed list", partition.getId());
         snapshotCompletedTablets.add(partition.getId());
         markSnapshotDoneOnServer(partition, offsetContext);
@@ -843,7 +843,7 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
     recordCheckpoint.setTime(Long.MAX_VALUE);
   }
 
-  private Boolean IsLastSnapshotRecordOfLastBatch(OpId opid) {
+  private Boolean isLastSnapshotRecordOfLastBatch(OpId opid) {
     String snapshotKey = ByteString.copyFrom(opid.getKey()).toStringUtf8();
     return (opid.getTerm() == Long.MAX_VALUE) &&
             (opid.getIndex() == Long.MAX_VALUE) &&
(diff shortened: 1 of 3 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -529,9 +546,13 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
...
 
+                OpId finalOpId = new OpId(resp.getTerm(), resp.getIndex(), resp.getKey(),
+                        resp.getWriteId(), resp.getSnapshotTime());
+                LOGGER.debug("Final OpId for tablet {} is {}", part.getId(), finalOpId);
+
                 // Process the response
-                for (CdcService.CDCSDKProtoRecordPB record :
-                        resp.getResp().getCdcSdkProtoRecordsList()) {
+                for (int idx = 0; idx < resp.getResp().getCdcSdkProtoRecordsList().size(); idx++) {
+                  CdcService.CDCSDKProtoRecordPB record = resp.getResp().getCdcSdkProtoRecordsList().get(idx);
```
vaibhav-yb, line 555: Another way to access the record at a given `idx` could be, it seems cleaner to me, but I'll let you pick, I am okay with both:

```
resp.getResp().getCdcSdkProtoRecords(idx);
```

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -593,6 +614,14 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
                           tId = YugabyteDBSchema.parseWithKeyspace(message.getTable(), connectorConfig.databaseName());
                         }
                         Objects.requireNonNull(tId);
+
+                        // if the DML record is the last snapshot record of the last snapshot batch, set all the fields
```
vaibhav-yb, line 618: Start the comment with a uppercase letter.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -500,7 +503,21 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
 
                 CdcSdkCheckpoint explicitCdcSdkCheckpoint = null;
                 if (taskContext.shouldEnableExplicitCheckpointing()) {
-                  explicitCdcSdkCheckpoint = tabletToExplicitCheckpoint.get(part.getId());
+                  CdcSdkCheckpoint checkpoint = tabletToExplicitCheckpoint.get(part.getId());
+
+                  // If the checkpoint in tabletToExplicitCheckpoint map corresponds to the last record's checkpoint
+                  // of last snapshot batch, it means kafka has consumed the last record. So, we should
+                  // not poll on this tablet anymore, instead mark snapshot done for this tablet.
+                  if(checkpoint != null && IsLastSnapshotRecordOfLastBatch(OpId.from(checkpoint))) {
```
vaibhav-yb, line 511: Add space after `if` --> `if (condition) {}`

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -593,6 +614,14 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
                           tId = YugabyteDBSchema.parseWithKeyspace(message.getTable(), connectorConfig.databaseName());
                         }
                         Objects.requireNonNull(tId);
+
+                        // if the DML record is the last snapshot record of the last snapshot batch, set all the fields
+                        // of Opid to max values and snapshot key to LAST_SNAPSHOT_RECORD.
+                        if(isSnapshotCompleteMarker(finalOpId) &&
```
vaibhav-yb, line 620: Space after if --> `if (condition) {}`

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -805,6 +830,28 @@ public void markSnapshotDoneOnServer(YBPartition partition, YugabyteDBOffsetCont
...
+   */
+  private void setIdentificationMarkerForLastSnapshotRecord(OpId recordCheckpoint) {
+    recordCheckpoint.setTerm(Long.MAX_VALUE);
+    recordCheckpoint.setIndex(Long.MAX_VALUE);
+    recordCheckpoint.setKey(LAST_SNAPSHOT_RECORD_KEY);
+    recordCheckpoint.setWriteId(Integer.MAX_VALUE);
+    recordCheckpoint.setTime(Long.MAX_VALUE);
+  }
+
+  private Boolean IsLastSnapshotRecordOfLastBatch(OpId opid) {
```
vaibhav-yb, line 846: We follow camel casing in Java, so the first letter of the method name would be in lower case i.e. `isLastSnapshotRecordOfLastBatch`

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -805,6 +830,28 @@ public void markSnapshotDoneOnServer(YBPartition partition, YugabyteDBOffsetCont
 
   }
 
+  /**
+   * Last snapshot record of the last snapshot batch will have a unique marker that will be used for indentification.
+   * Set all the fields of the record's Checkpoint to max values and set snapshot key to 'LAST_SNAPSHOT_RECORD'
+   * @param recordCheckpoint the record's checkpoint of the last snapshot record
+   */
+  private void setIdentificationMarkerForLastSnapshotRecord(OpId recordCheckpoint) {
+    recordCheckpoint.setTerm(Long.MAX_VALUE);
```
vaibhav-yb, line 839: Rather than using setters for the values, create a new `OpId` object and assign it to `recordCheckpoint` - this will ensure that the `OpId` object remains read-only consistently and we do not accidentally end up modifying the values.

src/main/java/io/debezium/connector/yugabytedb/connection/OpId.java
```diff
@@ -65,6 +66,26 @@ private static byte[] parseKey(String keyString) {
         return Base64.getDecoder().decode(keyString);
     }
 
+    public void setTerm(long term) {
```
vaibhav-yb, line 69: In my opinion, setters are not needed, creating a new object and assigning should do the same job. Refer to my other comment in `YugabyteDBSnapshotChangeEventSource`

vaibhav-yb (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
siddharth2411: This wont change the actual object `recordCheckpoint` is pointing to. We, ultimately, pass `recordCheckpoint` to kafka.

siddharth2411: This change has been done.

PR #306, opened 2023-12-11
Rectify logic for tablets to wait for callback in Snapshot flow

siddharth2411:
This PR is merged PR of [#302](https://github.com/yugabyte/debezium-connector-yugabytedb/pull/302) and #304. One PR wont work without the other.

Addresses two scenarios that finally leads to data loss:
1. Data loss in snapshot phase when connector restarts during consumption of the last snapshot batch yugabyte/yugabyte-db#20051
2. Data loss in streaming phase because connector is unable to stuck in snapshot phase if it restarts anytime after snapshot bootstrap is completed. yugabyte/yugabyte-db#20272

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -465,6 +465,8 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
 
       // Helper internal variable to log GetChanges request at regular intervals.
       long lastLoggedTimeForGetChanges = System.currentTimeMillis();
+      // Stores the READ records recevied in a single GetChanges call
+      long readRecordsReceived = 0;
 
       while (context.isRunning() && retryCount <= this.connectorConfig.maxConnectorRetries()) {
         try {
@@ -498,6 +500,8 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
                   continue;
                 }
 
+                // Reset for the upcoming getChanges call
+                readRecordsReceived = 0;
                 CdcSdkCheckpoint explicitCdcSdkCheckpoint = null;
                 if (taskContext.shouldEnableExplicitCheckpointing()) {
                   explicitCdcSdkCheckpoint = tabletToExplicitCheckpoint.get(part.getId());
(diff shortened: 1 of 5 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -498,9 +502,26 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
...
+
+                  // If the checkpoint in tabletToExplicitCheckpoint map corresponds to the last record's checkpoint
+                  // of last snapshot batch, it means kafka has consumed the last record. So, we should
+                  // not poll on this tablet anymore, instead mark snapshot done for this tablet.
+                  if (checkpoint != null && isLastSnapshotRecordOfLastBatch(OpId.from(checkpoint))) {
+                    LOGGER.debug("Already received the last record's modified checkpoint in kafka's callback. " +
+                            "Discontinuing polling on tablet {}", part.getId());
+                    LOGGER.info("Adding {} to the list of snapshot completed tablets", part.getId());
+                    snapshotCompletedTablets.add(part.getId());
+                    markSnapshotDoneOnServer(part, previousOffset);
```
suranjan, line 520: only when markSnapshotDoneOnServer is complete we should do others. markSnapshotDone can throw error.

suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
suranjan: Can you check the behaviour  if markSnapshotDoneOnServer fails?

siddharth2411: Checked the behavior - it can either fail to switch to streaming phase or succesfully switch to streaming depending on the number of tablets chosen for snapshot and the retry count. 

Based on our discussion, finally, changing it to this flow: First call `markSnapshotDoneOnServer()` and then only add the tablet to `snapshotCompletedTablets` set.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -778,6 +813,10 @@ public void markSnapshotDoneOnServer(YBPartition partition, YugabyteDBOffsetCont
     while (retryCount <= connectorConfig.maxConnectorRetries()) {
       try {
         LOGGER.info("Marking snapshot completed on service for table {} tablet {}", partition.getTableId(), partition.getTabletId());
+        // This flag is only meant for testing purposes only.
+        if(FAIL_WHEN_MARKING_SNAPSHOT_DONE) {
```
suranjan, line 817: formatting

suranjan (approved):

PR #307, opened 2023-12-12
Added test to repro data loss of last snapshot batch

siddharth2411:
### Problem
Consider this scenario:
Connector, in the snapshot phase, received the last batch of snapshot records. Since all the tablets were removed from the `shouldWaitForCallback` set right after the snapshot bootstrap call, connector will straight away send `snapshotDoneKey` even without waiting for Kafka to send callback for the last batch of snapshot records. Now if kafka had partially consumed the last batch of records and the connector restarted, the remaining records would be lost because connector when reading the cdc_state would conclude that snapshot has been completed on that particular tablet.  

### Solution
This is fixed by #306 

### Testing
Added new test `VerifyNoDataLossIfConnectorRestartDuringLastBatchConsumption` to reproduce the scenario. 

- Ran the test without fix in #306 and it fails when comparing the snapshot key before restarting the connector 

snapshot key of explicit checkpoint - some non-empty string
snapshot key in GetCheckpoint response - null (because snapshot was marked done on the server)

- Ran the test with the fix in #306 and both the snapshot keys are equal.

### Linked Issue
yugabyte/yugabyte-db#20272

### DIFF
--- a/src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotResumeTest.java
+++ b/src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotResumeTest.java
@@ -162,7 +162,6 @@ public class YugabyteDBSnapshotResumeTest extends YugabyteDBContainerTestBase {
 
 		// Wait for kafka to receive all the expected records before consuming records from kafka topic.
 		Awaitility.await()
-				.atLeast(Duration.ofSeconds(1))
 				.atMost(Duration.ofSeconds(180))
 				.until(() -> (getNonConsumedRecordCount() == totalRecordExpectedAfterPartialConsumption));
 
--- a/src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotResumeTest.java
+++ b/src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotResumeTest.java
@@ -8,6 +8,8 @@ import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.source.SourceRecord;
 import org.awaitility.Awaitility;
 import org.junit.jupiter.api.*;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.yb.client.CdcSdkCheckpoint;
 import org.yb.client.GetCheckpointResponse;
 import org.yb.client.YBClient;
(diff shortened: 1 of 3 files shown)


src/test/java/io/debezium/connector/yugabytedb/common/TestBaseClass.java
```diff
@@ -312,6 +381,50 @@ protected void waitAndFailIfCannotConsume(List<SourceRecord> records, long recor
...
 
+  protected void insertBulkRecords(int numRecords, String fullTableName) {
+    String formatInsertString = "INSERT INTO " + fullTableName + " VALUES (%d);";
+    TestHelper.executeBulk(formatInsertString, numRecords, DEFAULT_COLOCATED_DB_NAME);
+  }
+
+  /**
+   * Helper function to create the required tables in the database DEFAULT_COLOCATED_DB_NAME
+   */
+  protected void createTables(boolean colocation) {
```
vaibhav-yb, line 392: I think moving these methods out of the snapshot class won't make much sense since these tables are very specific to the snapshot tests only. Moving it to the `TestBaseClass` would mean that they can be used generically.

Maybe we can do one thing, rename the method to `createTablesInColocatedDB` and the next method to `dropTablesInColocatedDB` - that will ensure the method's name conveys what it is doing.

Similar renaming should be done for the insert methods as we have a method with a conflicting name in the class `TestHelper`.

vaibhav-yb (changes requested):

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotResumeTest.java
```diff
@@ -132,6 +132,78 @@ public void verifySnapshotIsResumedFromKey() throws Exception {
 		YugabyteDBSnapshotChangeEventSource.TRACK_EXPLICIT_CHECKPOINTS = false;
 	}
 
+	@ParameterizedTest
```
vaibhav-yb: nit: fix indentation

src/test/java/io/debezium/connector/yugabytedb/common/TestBaseClass.java
```diff
@@ -196,6 +200,71 @@ public void taskStarted() {
     });
   }
 
+  public int getNonConsumedRecordCount() {
```
vaibhav-yb, line 203: Add a javadoc comment for this method.

src/test/java/io/debezium/connector/yugabytedb/common/TestBaseClass.java
```diff
@@ -196,6 +200,71 @@ public void taskStarted() {
...
+    return linesConsumed.size();
+  }
+
+  public void startEngineWithPartialConsumptionOfLastBatch(Configuration.Builder configBuilder, int totalRecordsToConsume, DebeziumEngine.CompletionCallback callback) {
```
vaibhav-yb, line 207: Add javadoc comment

vaibhav-yb (approved):
Approved with a few minor comments.

PR #311, opened 2023-12-15
Fix flakiness in the test infra

vaibhav-yb:
## Problem

There are some related problems we are tracking here:
1. We do not have a cleaner way to set Gflags and start `yugabyted` inside the container. If we have to change/add one GFlag, we may need to make additional modifications to the test, for example, in `YugabyteDBTabletSplitTest`
2. Some tests are flaky and are not modified according to the consistent snapshot behaviour, as a result they end up failing at the assertion stages where we verify the count.

## Solution

This PR fixes the flakiness behaviour of the tests as well as introduces methods in the `TestBaseClass` i.e. `setMasterFlags` and `setTserverFlags` which will help simplify the process.

### Usage
The methods are to be used before the `initializeYBContainer()` method and if the container is started separately in the test then the flags need to be set before obtaining the container. For example:

* Inside beforeAll method

```java
public static void beforeAll() {
  setMasterFlags(masterFlags);
  setTserverFlags(tserverFlags);
  initializeYBContainer();
}
```

* Before obtaining the container

```java
  // test code
  setMasterFlags(masterFlags);
  setTserverFlags(tserverFlags);
  ybContainer = TestHelper.getYbContainer();
  // further test code
```


src/test/java/io/debezium/connector/yugabytedb/YugabyteDBExplicitCheckpointingTest.java
```diff
@@ -143,90 +143,4 @@ public void verifyCommitOffsetCheckpointAndGetCheckpointBehaviour(boolean consis
         // Stop the engine started in this test.
         engine.stop();
     }
-
-    @ParameterizedTest
-    @MethodSource("io.debezium.connector.yugabytedb.TestHelper#streamTypeProviderForStreaming")
```
vaibhav-yb, line 148: This test was redundant and took ~13 minutes to run so I have removed it to save some time.

suranjan (approved):

PR #326, opened 2024-02-15
[yugabyte/yugabyte-db#15718] Fix for data loss because of incorrect checkpoint advancing

vaibhav-yb:
## Problem

There were 2 bugs - one in the service where the safe time was being updated from the request safe time rather than the explicit checkpoint field and another in connector where we were advancing the explicit checkpoints incorrectly - that caused data loss if there was any connector restart during a certain scenario. The scenarios is as follows:

**Bug 1 (Service):**
1. Checkpoint in `cdc_state` is marked at `X1.Y1` and `t1`.
2. Connector requests for changes from checkpoint `X2.Y2` and safe time `t2`.
3. Service sends the changes and marks the safe time as `t2` - at this point, the connector crashes without publishing the data and restarts. Safe time will be marked as `t2` since the service will pick the safe time value from `GetChangesRequest` rather than the `explicit_checkpoint` field.
4. Upon restart, it will read the checkpoint from `cdc_state` as `X1.Y1` **but** with safe time as `t2` so whenever the service will read any record which has a `commit_time < t2`, that record will be filtered out and hence there will be data loss.

**Bug 2 (Connector):**
1. We call GetChanges with a from_op_id `<term:1,index:5,time:100>`
    a. A batch of 5 records is received and the `lastRecordCheckpoint` will be updated to `<term:1,index:5,key=keyForTheRecord,writeId=writeIdForTheRecord,time:99>`
    b. The explicitCheckpoint is at a lower value so the optimisation in picture will not be executed
    c. We received the callback on the last record published so far whose checkpoint would be `lastRecordCheckpoint` and explicitCheckpoint would become `<term:1,index:5,key=keyForTheRecord,writeId=writeIdForTheRecord,time:99>`
    d. `response.cdcsdkCheckpoint = <term:1,index:5,key=someValue,writeId=someValue,time=100>`
2. Next GetChanges will be called with a from_op_id `<term:1,index:5,key=someValue,writeId=someValue,time=100>`
    a. Another batch of 5 records is received and `lastRecordCheckpoint` will be updated to the value for the last record which will have a term = 1, index = 5 and time = 99
    b. Now since we are only comparing term, index and time values, it will become equal to the `explicitCheckpoint`
      i. This will end up advancing the explicit checkpoint to `response.cdcsdkCheckpoint`
    c. It is possible that the records are still in the connector queue and we haven’t published them so there will be no callback yet, but then we incremented the explicit checkpoint without a callback
3. Another `GetChanges` is called
    a. **This will end up marking the advanced explicitCheckpoint on the service**
4. There’s a connector restart
    b. Connector had not published the records in its queue but since the `explicitCheckpoint` was incremented, we will not get those records again, thus causing data loss

## Solution

1. The service side bug has been addressed by the commit https://github.com/yugabyte/yugabyte-db/commit/9246120dc5178cd38c31f24ad3ac7966910931d9
2. This PR has the connector fix to modify the method which checks whether or not to advance the explicit checkpoint.


src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
```diff
@@ -403,4 +404,93 @@ public void shouldWorkWithNullValues(boolean consistentSnapshot, boolean useSnap
...
+
+        String tabletId = tablets.iterator().next();
+        assertNotNull(tabletId);
+        assertEquals(1, tablets.size());
+
+        // This check will verify that we are getting some explicit checkpoint values, once that is confirmed, stop updating the explicit checkpoints.
+        Awaitility.await()
+          .atMost(Duration.ofMinutes(1))
+          .until(() -> YugabyteDBStreamingChangeEventSource.TEST_explicitCheckpoints.get(tabletId) != null);
+        YugabyteDBStreamingChangeEventSource.UPDATE_EXPLICIT_CHECKPOINT = false;
```
suranjan, line 447: How do we get explicit checkpoint without inserting anything?

suranjan (changes requested):

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
vaibhav-yb: If the response is empty i.e. has no records even then we keep moving the checkpoint forward with the response. Now the initial `GetChanges` calls give us a DDL and then the response keeps moving forward - the safe time value keeps moving forward, consequently the explicit checkpoint also has a value.

Code reference:
https://github.com/yugabyte/debezium-connector-yugabytedb/blob/main/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java#L762

src/main/java/io/debezium/connector/yugabytedb/connection/OpId.java
```diff
@@ -163,8 +163,17 @@ public CdcSdkCheckpoint toCdcSdkCheckpoint() {
      * the corresponding values in {@link CdcSdkCheckpoint}
      */
     public boolean isLesserThanOrEqualTo(CdcSdkCheckpoint checkpoint) {
-        return (checkpoint != null && this.term <= checkpoint.getTerm()
-                && this.index <= checkpoint.getIndex() && this.time <= checkpoint.getTime());
+        if (checkpoint == null) {
+            return false;
+        }
+
+        if (this.term < checkpoint.getTerm() || this.index < checkpoint.getIndex() || this.time < checkpoint.getTime()) {
```
Sumukh-Phalgaonkar, line 170: Should't we have `AND` in place of `OR` between the clauses here?

src/test/java/io/debezium/connector/yugabytedb/TestHelper.java
```diff
@@ -763,8 +763,8 @@ private static YugabyteDBValueConverterBuilder getPostgresValueConverterBuilder(
 
     public static Stream<Arguments> streamTypeProviderForStreaming() {
         return Stream.of(
-                Arguments.of(false, false), // Older stream
-                Arguments.of(true, false)); // NO_EXPORT stream
+                Arguments.of(false, false)); // Older stream
+//                Arguments.of(true, false)); // NO_EXPORT stream
```
Sumukh-Phalgaonkar, line 767: Reminder: Uncomment before merging

src/main/java/io/debezium/connector/yugabytedb/connection/OpId.java
vaibhav-yb: Suppose we have a case where we received the finalOpId as `1.2.[].0.100` and we genuinely do not have anything further in the WAL, not even a `NO_OP`, in that case our time in the response checkpoint will keep on increasing(`1.2.[].0.101`, `1.2.[].0.102`, `1.2.[].0.103` and so on) without the `term` or `index` moving ahead - we would want to increment the advanced checkpoint even when the only increment is in time.

If we add an `AND` here instead of `OR`, we will end up failing the check and won't increment the checkpoint contrary to what we are trying to achieve here.

src/main/java/io/debezium/connector/yugabytedb/connection/OpId.java
```diff
@@ -163,8 +163,17 @@ public CdcSdkCheckpoint toCdcSdkCheckpoint() {
...
+            return true;
+        } else if (this.term == checkpoint.getTerm() && this.index == checkpoint.getIndex() && this.time == checkpoint.getTime()) {
+            return Arrays.equals(this.key, checkpoint.getKey()) && this.write_id == checkpoint.getWriteId();
```
suranjan, line 173: return this.write_id <= checkpoint.getWriteId();

suranjan (changes requested):

suranjan (approved):

PR #328, opened 2024-03-01
Use configuration to log every GetChangesRequest

vaibhav-yb:
This PR leverages an existing and unused configuration property `log.get.changes` to evaluate whether the connector should log every `GetChangesRequest` going to the service, the type and default behaviour is as follows:

`log.get.changes` = `BOOLEAN`

1. If `log.get.changes = false` (default):
    a. If debug logging is enabled then every `GetChangesRequest` will be logged
    b. If connector is being run at the `INFO` logging level then `GetChangesRequest` will be logged at an interval of 5 minutes
2. If `log.get.changes = true`
    a. Every `GetChangesRequest` will be logged

Additionally, this PR modifies the log statement in all the `ChangeEventSource` implementations so that they are all uniform and follow the format:

```
Requesting changes for table <tableName> tablet <tabletId>, explicit_checkpoint: 1.3.[].0.7001191540353540096 from_op_id: 1:3:[]:0:7001191540353540096
```

The statement in case of snapshot would follow the same format except for the word `changes` being replaced with `snapshot`:

```
Requesting snapshot for table <tableName> tablet <tabletId>, explicit_checkpoint: 1.3.[].0.7001191540353540096 from_op_id: 1:3:[]:0:7001191540353540096
```

One final change in this PR is that the method `OpId#getKeyString` now returns an array representation of the key which aligns with the representation of the key in the `CdcsdkCheckpoint` class in `yb-client`.


src/main/java/io/debezium/connector/yugabytedb/connection/OpId.java
```diff
@@ -54,7 +54,7 @@ public String getKeyString() {
             return "null";
         }
 
-        return Base64.getEncoder().encodeToString(key);
+        return Arrays.toString(key);
```
suranjan, line 57: why this change?

suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/connection/OpId.java
vaibhav-yb: This is to bring the `toString` implementations of `OpId` and `CdcsdkCheckpoint` aligned - the `CdcdkCheckpoint#toString` uses this array like representation but over here if we directly printed the encoded string, it becomes difficult for us to verify keys' equality while reading the logs so that's why this change.

suranjan (approved):

src/main/java/io/debezium/connector/yugabytedb/connection/OpId.java
suranjan: Don't remember but maybe we did it because of security reasons

PR #331, opened 2024-03-07
[yugabyte/yugabyte-db#20431] Publish last snapshot record again if callback not received

vaibhav-yb:
## Problem

1. Kafka topic is created with multiple partitions (2 in this example), let’s call them `partition_0` and `partition_1`
2. Connector takes snapshot
3. When the connector receives the last record of the last snapshot batch, it modifies its offset to `Long.MAX_VALUE` so that when Kafka executes a callback on this, we know that Kafka has received the last record of the snapshot process and we can transition to streaming.
4. But here’s the catch, suppose we have the second last record of the snapshot batch which has the offset `offset_second_last_record` and the last record’s modified offset is `offset_last_record`
5. Now even when the connector reads it in the correct order i.e. `second_last` and then `last`, since there are two partitions, there seems to be some kind of situation where the `second_last` record is published later in the topic and that is where the callback gets stuck since we will never receive the callback on the `last` record now.

The above was leading to a problem where we were never transitioning out of snapshot phase thus leading to a task getting stuck there.

## Solution

This PR implements the solution by introducing a timeout `last.callback.timeout.ms` where the logic is to publish the last snapshot record again which will ensure that we receive a callback on the last record of last snapshot batch and gracefully move ahead.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -494,6 +494,12 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
                   if (!snapshotCompletedTablets.contains(part.getId()) && hasCallbackTimeoutExceeded(part)) {
                     LOGGER.info("Publishing last snapshot record for partition {} again", part.getId());
                     publishLastSnapshotRecord(part, previousOffset);
+
+                    // Also update the last GetChanges time to ensure that we do not end up publishing
+                    // the last record continuously without exhausting the delay. In other words, it
+                    // can be understood that the publishing the last snapshot record was the result
+                    // of a GetChanges call, thus we are updating the map for time.
+                    lastGetChangesTime.put(part.getId(), System.currentTimeMillis());
                   }
 
                   continue;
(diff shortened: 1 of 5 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -480,7 +482,16 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
                   if (!snapshotCompletedTablets.contains(part.getId()) && taskContext.shouldEnableExplicitCheckpointing()) {
                     doSnapshotCompletionCheck(part, snapshotCompletedTablets, tabletsWaitingForCallback, previousOffset);
                   }
-                  continue;
+
+                  if (!snapshotCompletedTablets.contains(part.getId()) && isCallbackTimeoutExceeded(part)) {
```
suranjan, line 486: Do we handle empty tablets?

vaibhav-yb: Yes, for empty tablets this flow will never be hit as `snapshotCompletedTablets` will contain the partition and the condition `!snapshotCompletedTablets.contains(part.getId())` will become `false`.

PR #335, opened 2024-06-11
[WIP] Create a JDBC connection only while decoding array types

vaibhav-yb:
## Problem

The `YugabyteDBConnector` opens a global JDBC connection which is used to decode data types, this connection is created only once per task and is retained for the complete lifetime of the task. However, one task per connection creates a problem when the task count is high and we end up in a state where we have a lot of JDBC connections open.

Since the change events the connector receives are already decoded in case of non-array types, a JDBC connection is not really needed to decode them so unless there's an array type in the table the connector is polling, the connection keeps hogging resources.

## Solution

This PR aims to change the behaviour and only uses a JDBC connection while decoding array types.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceFactory.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceFactory.java
@@ -32,7 +32,6 @@ public class YugabyteDBChangeEventSourceFactory implements ChangeEventSourceFact
     private static final Logger LOGGER = LoggerFactory.getLogger(YugabyteDBChangeEventSourceFactory.class);
 
     private final YugabyteDBConnectorConfig configuration;
-    private final YugabyteDBConnection jdbcConnection;
     private final ErrorHandler errorHandler;
     private final YugabyteDBEventDispatcher<TableId> dispatcher;
     private final Clock clock;
@@ -46,7 +45,6 @@ public class YugabyteDBChangeEventSourceFactory implements ChangeEventSourceFact
 
     public YugabyteDBChangeEventSourceFactory(YugabyteDBConnectorConfig configuration,
                                               Snapshotter snapshotter,
-                                              YugabyteDBConnection jdbcConnection,
                                               ErrorHandler errorHandler,
                                               YugabyteDBEventDispatcher<TableId> dispatcher,
                                               Clock clock, YugabyteDBSchema schema,
@@ -56,7 +54,6 @@ public class YugabyteDBChangeEventSourceFactory implements ChangeEventSourceFact
                                               SlotState startingSlotInfo,
                                               ChangeEventQueue<DataChangeEvent> queue) {
         this.configuration = configuration;
-        this.jdbcConnection = jdbcConnection;
         this.errorHandler = errorHandler;
         this.dispatcher = dispatcher;
         this.clock = clock;
@@ -76,7 +73,6 @@ public class YugabyteDBChangeEventSourceFactory implements ChangeEventSourceFact
                 configuration,
                 taskContext,
                 snapshotter,
-                jdbcConnection,
                 schema,
                 dispatcher,
                 clock,
@@ -91,7 +87,6 @@ public class YugabyteDBChangeEventSourceFactory implements ChangeEventSourceFact
             return new YugabyteDBStreamingChangeEventSource(
                     configuration,
                     snapshotter,
-                    jdbcConnection,
                     dispatcher,
                     errorHandler,
                     clock,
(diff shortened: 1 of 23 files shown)


src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
```diff
@@ -29,7 +30,7 @@
  * @author Vaibhav Kushwaha (vkushwaha@yugabyte.com)
  */
 
-public class YugabyteDBDatatypesTest extends YugabyteDBContainerTestBase {
+public class YugabyteDBDatatypesTest extends YugabytedTestBase {
```
vaibhav-yb, line 33: Todo Vaibhav: Revert this change before final review.

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBCompleteTypesTest.java
```diff
@@ -26,16 +31,17 @@
 
 import static org.junit.jupiter.api.Assertions.*;
 
-public class YugabyteDBCompleteTypesTest extends YugabyteDBContainerTestBase {
+public class YugabyteDBCompleteTypesTest extends YugabytedTestBase {
```
vaibhav-yb, line 34: Todo Vaibhav: Revert this change before final review.

src/test/java/io/debezium/connector/yugabytedb/TestHelper.java
```diff
@@ -758,13 +758,40 @@ private static YugabyteDBValueConverterBuilder getPostgresValueConverterBuilder(
 
     public static Stream<Arguments> streamTypeProviderForStreaming() {
         return Stream.of(
-                Arguments.of(false, false), // Older stream
-                Arguments.of(true, false)); // NO_EXPORT stream
+                Arguments.of(false, false)); // Older stream
+//                Arguments.of(true, false)); // NO_EXPORT stream
```
vaibhav-yb, line 762: Todo Vaibhav: Revert this change before final review.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceFactory.java
```diff
@@ -46,7 +45,6 @@ public class YugabyteDBChangeEventSourceFactory implements ChangeEventSourceFact
 
     public YugabyteDBChangeEventSourceFactory(YugabyteDBConnectorConfig configuration,
                                               Snapshotter snapshotter,
-                                              YugabyteDBConnection jdbcConnection,
```
suranjan, line 49: why change the signature?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceFactory.java
```diff
@@ -119,14 +113,6 @@ public StreamingChangeEventSource<YBPartition, YugabyteDBOffsetContext> getStrea
     public Optional<IncrementalSnapshotChangeEventSource<YBPartition, ? extends DataCollectionId>> getIncrementalSnapshotChangeEventSource(YugabyteDBOffsetContext offsetContext,
                                                                                                                               SnapshotProgressListener snapshotProgressListener,
                                                                                                                               DataChangeEventListener dataChangeEventListener) {
-        final SignalBasedIncrementalSnapshotChangeEventSource<YBPartition, TableId> incrementalSnapshotChangeEventSource = new SignalBasedIncrementalSnapshotChangeEventSource<YBPartition, TableId>(
```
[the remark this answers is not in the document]
suranjan: avoid formatting change

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeRecordEmitter.java
```diff
@@ -196,8 +196,7 @@ private Object[] columnValues(List<ReplicationMessage.Column> columns, TableId t
             undeliveredToastableColumns.remove(columnName);
             int position = getPosition(columnName, table, values);
             if (position != -1) {
-                Object value = column.getValue(() -> (BaseConnection) connection.connection(),
-                        connectorConfig.includeUnknownDatatypes());
```
suranjan, line 200: why is this change needed,  connection.connection() can return null/connection depending on the context

suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeRecordEmitter.java
vaibhav-yb: The `YugabyteDBConnection` object doesn't have any context of what type of data it will be used to decode. In case an array type is encountered, then a JDBC connection is created to be used otherwise it would be just the `YugabyteDBConnection` object.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceFactory.java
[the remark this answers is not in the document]
vaibhav-yb: Wasn't intentional, some wrong config with IDE, will fix this.

PR #345, opened 2024-08-12
[DBZ] Read offset from the topic to mark explicit checkpoints

vaibhav-yb:


src/main/java/io/debezium/connector/yugabytedb/YBPartition.java
```diff
@@ -42,6 +43,11 @@ public YBPartition(String tableId, String tabletId, boolean isTableColocated) {
         this.colocated = isTableColocated;
     }
 
+    public static YBPartition from(String fullPartitionId) {
+        String[] tableTablet = fullPartitionId.split("\\.");
+        return new YBPartition(tableTablet[0], tableTablet[1], true);
```
vaibhav-yb, line 48: This currently assumed that the partition ID in the topic will always be the fullPartitionId but that will not be the case if connector is in streaming mode.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
```diff
@@ -363,6 +378,87 @@ protected Iterable<Field> getAllConfigurationFields() {
...
+
+     The records are also published in the same order i.e. tablet_0, tablet_1, tablet_2
+
+     But it is not guaranteed that while reading the offsets from kafka we will read in the same order, we can end up reading the partitions/tablets in the order
+     tablet_2
+     tablet_3
+     tablet_1
+
+     and if we call commitOffset on each of the partition, we will basically be overriding the offsets with a lower value which we do not want to happen
+     */
```
vaibhav-yb, line 400: Clean this comment.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -492,14 +492,14 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
                   // If the timeout has exceeded from the last GetChanges call and we haven't received
                   // any callback on the last snapshot record yet, publish last snapshot record again.
                   if (!snapshotCompletedTablets.contains(part.getId()) && hasCallbackTimeoutExceeded(part)) {
-                    LOGGER.info("Publishing last snapshot record for partition {} again", part.getId());
-                    publishLastSnapshotRecord(part, previousOffset);
+                    LOGGER.info("[LOG ONLY] Publishing last snapshot record for partition {} again", part.getId());
```
vaibhav-yb, line 495: This logic can be removed now.

src/main/java/io/debezium/connector/yugabytedb/connection/HashPartition.java
```diff
@@ -232,7 +232,7 @@ public String toString() {
   }
 
   public YBPartition toYBPartition() {
-    return new YBPartition(tableId, tabletId);
+    return new YBPartition(tableId, tabletId, true);
```
vaibhav-yb, line 235: Is there a way to determine at this stage whether a table should be colocated?

src/main/java/io/debezium/connector/yugabytedb/connection/OpId.java
```diff
@@ -126,12 +126,18 @@ public int hashCode() {
     @Override
     public int compareTo(OpId o) {
         // Unsigned comparison
-        if (term != o.term)
-            return term + Long.MIN_VALUE < o.term + Long.MIN_VALUE ? -1 : 1;
-        else if (index != o.index)
-            return index + Long.MIN_VALUE < o.index + Long.MIN_VALUE ? -1 : 1;
-        else
-            return write_id + Long.MIN_VALUE < o.write_id + Long.MIN_VALUE ? -1 : 1;
+        if (term != o.term) {
```
vaibhav-yb, line 129: This can be reverted since we are using the `isLesserThanOrEqualTo` check to ensure that the logic to compare the checkpoints remain the same everywhere.

vaibhav-yb:
This PR is being closed in lieu of #347, #348 and #351

PR #346, opened 2024-08-13
[DBZ] Use full partition ID for colocated tables

vaibhav-yb:
## Problem

Currently, the `YBPartition` class object returns a partition ID by the method `getId()` which depends whether the table is in colocated:
- colocated: `tableId.tabletId`
- non colocated: `tabletId`

So all the offsets, keys, OpIds, etc. are stored against this ID.

Now since the service had special handling for snapshot for colocated tables, we treat every `YBPartition` as a colocated object in the snapshot phase - what that means is that we can get the snapshot `GetChangesResponse` for each colocated table individually. While we treat every `YBPartition` in streaming phase as non-colocated since we do not have a way to get the changes for each tablet separately.

This creates a problem at the management level that we have to take care of properly creating the `YBPartition` object otherwise we'll end up returning a wrong ID if we use `YBPartition#getId` method.

## Solution

This PR removes the colocated property from the `YBPartition` class and introduces the following changes
- `YBPartition#getID` will now return full partition ID every time, this makes the ID a consistent value across snapshot and streaming phase.
- If a table is colocated and we're in the streaming phase, we will update the offset values for all the tables belonging to the colocated tablet upon every `GetChanges` call.
- The `tabletSafeTime` map is now moved to `YugabyteDBOffsetContext` so that it can be handled consistently for all the change event source classes.

### Test plan

Ran existing tests for colocated tables as well as non colocated tables to ensure this PR breaks nothing.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YBPartition.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YBPartition.java
@@ -26,20 +26,9 @@ public class YBPartition implements Partition {
     private final String tabletId;
     private final String tableId;
 
-    private boolean colocated;
-
     public YBPartition(String tableId, String tabletId) {
         this.tableId = tableId;
         this.tabletId = tabletId;
-
-        // By default, assume that the table is not colocated.
-        this.colocated = false;
-    }
-
-    public YBPartition(String tableId, String tabletId, boolean isTableColocated) {
-        this.tableId = tableId;
-        this.tabletId = tabletId;
-        this.colocated = isTableColocated;
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


src/main/java/io/debezium/connector/yugabytedb/YBPartition.java
```diff
@@ -26,20 +26,9 @@ public class YBPartition implements Partition {
     private final String tabletId;
     private final String tableId;
 
-    private boolean colocated;
```
vaibhav-yb, line 29: Removing the colocated property completely to prevent any accidental usage.

PR #348, opened 2024-08-14
[DBZ] Read offset from Kafka for every commit callback

vaibhav-yb:
# Note: This PR breaks explicit checkpointing in the tablet splitting case, it will be fixed in a follow-up PR.

## Problem

There are two issues with the current checkpointing mechanism using the callbacks:
* https://issues.redhat.com/browse/DBZ-6026
* https://issues.redhat.com/browse/DBZ-7816

They both are potential cause of data loss and have been reproduced manually as well.

**DBZ-6026 -**

The current code uses a method [BaseSourceTask#logStatistics](https://github.com/debezium/debezium/blob/v1.9.5.Final/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java#L178) to log some information, however, it also ends up updating the offset map which is then used for a callback.

However, with this, there's a possibility that if there's a `commit()` callback after the statistics are logged and before the records are returned from the [BaseSourceTask#poll](https://github.com/debezium/debezium/blob/v1.9.5.Final/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java#L162), it will end up marking the checkpoint on service but if the connector restarts in that window, there will be a data loss.

*Steps to reproduce:*
1. Put a sleep of 2 minutes after `logStatistics()` is called and before records are returned from [BaseSourceTask#poll](https://github.com/debezium/debezium/blob/v1.9.5.Final/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java#L162).
    a. Add a log so that we see when the method is getting called.
2. Create a table with a single tablet.
3. Insert a record and wait for the log in 1a to appear, that will indicate that records are not yet returned.
4. Wait for commit callback - this generally comes before the sleep gets over.
5. Restart the connector.
6. Upon restart, the connector will start from the checkpoint in step 4 and the record inserted in step 3 will never be streamed.

**DBZ-7816 -**

According to [Kafka docs](https://github.com/apache/kafka/blob/864744ffd4ddc3b0d216a3049ee0c61e9c0d3ad1/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L901), callbacks for the same Kafka partition are guaranteed to be in order but callbacks for different Kafka partitions can come out of order which can lead to a potential data loss window as mentioned in [DBZ-7816](https://issues.redhat.com/browse/DBZ-7816)

*Steps to reproduce:*
1. Create a table with 20 tablets and fill it with 100k records.
2. Create a topic with 50 partitions.
3. Start snapshot on the table with `snapshot.mode=initial`
4. Once the snapshot is finished, the tablet will be added to wait list and if the callbacks come out of order, it is possible that we will never receive a callback for the last snapshot record and we will never transition from snapshot to streaming.

Note that out of all the experiments performed for this issue, we were able to reproduce it 100% of the times but there's a possibility that it might not reproduce if callbacks are in order (1 out of 10 times maybe?)

## Solution

This PR includes the changes to override the `commit()` method in `YugabyteDBConnectorTask` and reads the offsets from Kafka partitions and uses the same offsets to send via commit callbacks so that they can be marked on service for checkpointing.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceCoordinator.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceCoordinator.java
@@ -127,8 +127,7 @@ public class YugabyteDBChangeEventSourceCoordinator extends ChangeEventSourceCoo
         previousLogContext.set(taskContext.configureLoggingContext(
             String.format("streaming|%s", taskContext.getTaskId())));
 
-        ((YugabyteDBTaskContext) taskContext).markSnapshotOver();
-        LOGGER.info("Snapshot flag upon transition: {}", ((YugabyteDBTaskContext) taskContext).isSnapshotInProgress());
+        LOGGER.info("Snapshot flag upon transition for task {}: {}", taskContext.getTaskId(), isSnapshotInProgress());
 
         for (Map.Entry<YBPartition, YugabyteDBOffsetContext> entry :
                 streamingOffsets.getOffsets().entrySet()) {
@@ -170,6 +169,10 @@ public class YugabyteDBChangeEventSourceCoordinator extends ChangeEventSourceCoo
         if (!commitOffsetLock.isLocked() && streamingSource != null && offset != null) {
             streamingSource.commitOffset(offset);
         }
+    }
+
+    protected boolean isSnapshotInProgress() {
+        return snapshotter.shouldSnapshot() && !snapshotSource.isSnapshotComplete();
     }
 
     private void setSnapshotStartLsn(YugabyteDBSnapshotChangeEventSource snapshotSource,
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceCoordinator.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceCoordinator.java
@@ -127,6 +127,9 @@ public class YugabyteDBChangeEventSourceCoordinator extends ChangeEventSourceCoo
         previousLogContext.set(taskContext.configureLoggingContext(
             String.format("streaming|%s", taskContext.getTaskId())));
 
+        ((YugabyteDBTaskContext) taskContext).markSnapshotOver();
+        LOGGER.info("Snapshot flag upon transition: {}", ((YugabyteDBTaskContext) taskContext).isSnapshotInProgress());
+
         for (Map.Entry<YBPartition, YugabyteDBOffsetContext> entry :
                 streamingOffsets.getOffsets().entrySet()) {
             initStreamEvents(entry.getKey(), entry.getValue());
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceCoordinator.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceCoordinator.java
@@ -155,7 +155,7 @@ public class YugabyteDBChangeEventSourceCoordinator extends ChangeEventSourceCoo
 
     @Override
     public void commitOffset(Map<String, ?> offset) {
-        if (snapshotSource == null) {
+        if (this.snapshotSource == null) {
             return;
         }
 
@@ -163,7 +163,7 @@ public class YugabyteDBChangeEventSourceCoordinator extends ChangeEventSourceCoo
         // streaming source only. If snapshot is complete, even then the callback should go to the
         // streaming source as in case of a finished snapshot, we do not want to do a duplicate call
         // for commitOffset.
-        if (!commitOffsetLock.isLocked() && snapshotter.shouldSnapshot() && !snapshotSource.isSnapshotComplete()) {
+        if (!commitOffsetLock.isLocked() && snapshotter.shouldSnapshot() && !this.snapshotSource.isSnapshotComplete()) {
             snapshotSource.commitOffset(offset);
             return;
         }
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceCoordinator.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceCoordinator.java
@@ -169,6 +169,9 @@ public class YugabyteDBChangeEventSourceCoordinator extends ChangeEventSourceCoo
         }
     }
 
+    /**
+     * @return true if the connector is in snapshot phase, false otherwise
+     */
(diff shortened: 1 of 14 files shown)


src/test/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotTest.java
```diff
@@ -1024,10 +1024,10 @@ private void verifyRecordCount(long recordsCount) {
...
     static Stream<Arguments> streamTypeProviderForSnapshotWithColocation() {
         return Stream.of(
-                Arguments.of(false, false, true), // Older stream with colocation
-                Arguments.of(false, false, false), // Older stream without colocation
-                Arguments.of(true, true, true), // USE_SNAPSHOT stream with colocation
-                Arguments.of(true, true, false));  // USE_SNAPSHOT stream without colocation
+                // Arguments.of(false, false, true), // Older stream with colocation
+                Arguments.of(false, false, false)); // Older stream without colocation
+                // Arguments.of(true, true, true), // USE_SNAPSHOT stream with colocation
+                // Arguments.of(true, true, false));  // USE_SNAPSHOT stream without colocation
```
vaibhav-yb, line 1030: This change will be reverted/removed.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -940,7 +940,7 @@ public void commitOffset(Map<String, ?> offset) {
                     // than one already present would throw the error: CDCSDK: Trying to fetch already GCed intents
                     if (this.tabletToExplicitCheckpoint.get(entry.getKey()) != null &&
                             tempOpId.getIndex() < this.tabletToExplicitCheckpoint.get(entry.getKey()).getIndex()) {
-                        LOGGER.warn("The received OpId {} is less than the older checkpoint {} for tablet {}",
+                        LOGGER.debug("The received OpId {} is less than the older checkpoint {} for tablet {}",
```
vaibhav-yb, line 943: This is converted to `debug` level considering the following scenario:
1. Last record published by the connector has OpId `1.5`
2. Connector received a callback on the above OpId so explicit checkpoint is also at `1.5`
3. Now there were couple of NO_OP on service so connector received empty batches and explicit checkpoint was advanced to `1.8`
4. Assume that for sometime there are no records, or there are no records after the record in step 1
5. This will result in a commit callback with checkpoint `1.5` since that was the last published record
6. Subsequently, we will always end up printing this warning log which could be confusing to users.

suranjan (approved):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceCoordinator.java
```diff
@@ -155,6 +155,10 @@ protected void executeChangeEventSources(CdcSourceTaskContext taskContext,
 
     @Override
     public void commitOffset(Map<String, ?> offset) {
+        if (snapshotSource == null) {
```
suranjan, line 158: if not local variable use `this`. Should follow that convention everywhere.
That way it becomes easy to see if null check is required or not.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
```diff
@@ -363,6 +382,110 @@ protected Iterable<Field> getAllConfigurationFields() {
...
+     * <code>{a=3,b=1,c=6}</code> then the value returned will be <code>{a=3,b=12,c=6}</code>.
+     * @param offsets the offset map read from Kafka topic
+     * @return a map with the values higher among the cached ybOffset and passed offsets map
+     */
+    protected Map<String, ?> getHigherOffsets(Map<String, ?> offsets) {
+        if (this.ybOffset == null) {
+            return offsets;
+        }
+
+        Map<String, String> finalOffsets = new HashMap<>();
```
suranjan, line 454: move it below the if block

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
```diff
@@ -363,6 +382,110 @@ protected Iterable<Field> getAllConfigurationFields() {
...
+        return finalOffsets;
+    }
+
+    /**
+     * @return true if the {@link YugabyteDBChangeEventSourceCoordinator} for this task has started
+     * executing the streaming source, false otherwise. In other words, this method indicates the
+     * status whether this task is in the snapshot phase.
+     */
+    protected boolean isTaskInSnapshotPhase() {
+        return coordinator.isSnapshotInProgress();
```
suranjan, line 486: use this and null check..if not initialized in the constructor

suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
```diff
@@ -64,6 +66,15 @@ public class YugabyteDBConnectorTask
...
     private volatile YugabyteDBSchema schema;
 
+    private YugabyteDBChangeEventSourceCoordinator coordinator;
+
+    private YBPartition.Provider partitionProvider;
+    private YugabyteDBOffsetContext.Loader offsetContextLoader;
+
+    private final ReentrantLock commitLock = new ReentrantLock();
+
+    protected volatile Map<String, ?> ybOffset;
```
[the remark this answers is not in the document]
vaibhav-yb: I have verified that upon restart (either connector or task), a new object for the task is created and no objects are shared i.e. everything gets created again.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
```diff
@@ -272,12 +285,15 @@ Offsets<YBPartition, YugabyteDBOffsetContext> getPreviousOffsetsFromProviderAndL
...
         boolean found = false;
-        for (YBPartition partition : partitions) {
-            YugabyteDBOffsetContext offset = offsets.getOffsets().get(partition); //offsets.get(partition);
-
-            if (offset != null) {
-                found = true;
-                LOGGER.info("Found previous partition offset {}: {}", partition, offset);
+        if (offsets != null) {
+            found = true;
+
```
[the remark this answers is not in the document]
vaibhav-yb: Making the log changes started giving OOM issues, it could be related so I am planning to introduce those changes separately in another PR later so that we keep things isolated.

suranjan (approved):

PR #351, opened 2024-09-02
[DBZ] Read offsets from Kafka with tablet splitting support

vaibhav-yb:
## Problem

With PR #351 getting merged, we are now reading offsets from the Kafka topic for the last published record for each source partition and using that offset for explicit checkpointing. The issue here is that to read from Kafka, we use the `Provider` class to give us a list of source partitions we need to read from, but the provider only reads from the static values passed on to the task in the configuration - this will cause issues with tablet splitting where if a tablet gets split and we start polling the children tablets, the provider will still return the parent partition and we will never be able to read the offsets for the children tablets.

## Solution

This PR adds the solution for the above mentioned problem by adding logic to read the list of partitions from the streaming change event source since that will have the access to the active partitions (including the children tablets) and hence we now have a way to get the partitions dynamically.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceCoordinator.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceCoordinator.java
@@ -135,7 +135,6 @@ public class YugabyteDBChangeEventSourceCoordinator extends ChangeEventSourceCoo
             initStreamEvents(entry.getKey(), entry.getValue());
         }
 
-        this.streamingChangeEventSource = (YugabyteDBStreamingChangeEventSource) streamingSource;
         LOGGER.info("Performing the streaming process now");
 
         while (context.isRunning()) {
@@ -158,6 +157,18 @@ public class YugabyteDBChangeEventSourceCoordinator extends ChangeEventSourceCoo
     }
 
     @Override
+    protected void streamEvents(ChangeEventSourceContext context, YBPartition partition,
+            YugabyteDBOffsetContext offsetContext) throws InterruptedException {
+        initStreamEvents(partition, offsetContext);
+        LOGGER.info("Starting streaming");
+
+        this.streamingChangeEventSource = (YugabyteDBStreamingChangeEventSource) streamingSource;
+
+        streamingSource.execute(context, partition, offsetContext);
+        LOGGER.info("Finished streaming");
+    }
+
+    @Override
     public void commitOffset(Map<String, ?> offset) {
         // Check if snapshotter is enabled, if it is not then callback should go to the
         // streaming source only. If snapshot is complete, even then the callback should go to the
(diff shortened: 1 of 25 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -101,6 +101,10 @@ public class YugabyteDBStreamingChangeEventSource implements
     protected Set<String> splitTabletsWaitingForCallback;
     protected List<HashPartition> partitionRanges;
 
+    // This tabletPairList has Pair<String, String> objects wherein the key is the table UUID
+    // and the value is tablet UUID
+    protected List<Pair<String, String>> tabletPairList;
```
vaibhav-yb, line 106: Take a lock or use copy-on-write list

vaibhav-yb: Or some sort of synchronous list - something which works across threads.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBChangeEventSourceCoordinator.java
```diff
@@ -173,6 +188,34 @@ public void commitOffset(Map<String, ?> offset) {
...
+     * which should be handled by the caller of this method.
+     */
+    public Optional<Set<YBPartition>> getPartitions() {
+        if (this.streamingChangeEventSource == null) {
+            LOGGER.info("Streaming change event source is null, returning empty value");
+            return Optional.empty();
+        }
+
+        Optional<Set<YBPartition>> ybPartitions =
+          Optional.of(this.streamingChangeEventSource.getActivePartitionsBeingPolled());
```
vaibhav-yb, line 204: this.streamingSource == null || (this.stteamingSource.getActivePartitionsbegingPolled().size() == 0)

- check if there's an empty() method for set.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
```diff
@@ -274,8 +274,33 @@ Map<YBPartition, YugabyteDBOffsetContext> getPreviousOffsetss(
...
+             snapshot phase. See YugabyteDBChangeEventCoordinator#getPartitions for more details.
+ 
+           Streaming phase:
+             We can rely on coordinator to leverage the streaming change event source and send us
+             the set of active partitions being polled which includes the dynamically created
+             partitions because of tablet splitting too. There can be a small window of time when
+             the partition list in streaming change event source will be empty and not populated,
+             so using the provider to get partitions is fine in this case as well.
+         */    
+        Optional<Set<YBPartition>> ybPartitions =
```
suranjan, line 300: If we know snapshot phase and streaming phase then lets separate out the logic

suranjan (changes requested):

suranjan (approved):

PR #353, opened 2024-09-03
[DBZ] Bug fix for snapshot not transitioning to streaming

vaibhav-yb:
## Problem

There are two bugs which were introduced when #348 and #352 were merged.

1. Wrong logic for the method `YugabyteDBConnectorTask#isTaskInSnapshotPhase`

```
protected boolean isTaskInSnapshotPhase() {
    return (this.coordinator == null) && this.coordinator.isSnapshotInProgress();
}
```

The above logic will always return `false` indicating that the task is never in snapshot phase.

2. The method `YugabyteDBSnapshotChangeEventSource#getExplicitCheckpoint` does not consider the checkpoint for last record of last snapshot batch.

## Solution

1. This PR fixes the method `YugabyteDBConnectorTask#isTaskInSnapshotPhase` by modifying the condition to use an `OR` operator.

2. For the method `YugabyteDBSnapshotChangeEventSource#getExplicitCheckpoint`, the condition is modified to check if the explicit checkpoint belongs to the last record of last snapshot batch then we should use that checkpoint.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
@@ -453,7 +453,8 @@ public class YugabyteDBConnectorTask
 
                         if (LOGGER.isDebugEnabled()) {
                             for (Map.Entry<String, ?> entry : ybOffset.entrySet()) {
-                                LOGGER.debug("Committing offset {} for partition {}", entry.getValue(), entry.getKey());
+                                LOGGER.debug("{} | Committing offset {} for partition {}",
+                                             taskContext.getTaskId(), entry.getValue(), entry.getKey());
                             }
                         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
```diff
@@ -514,7 +515,7 @@ public void commit() throws InterruptedException {
      * status whether this task is in the snapshot phase.
      */
     protected boolean isTaskInSnapshotPhase() {
-        return (this.coordinator == null) && this.coordinator.isSnapshotInProgress();
+        return (this.coordinator == null) || this.coordinator.isSnapshotInProgress();
```
suranjan, line 518: it will return true if coordinator is null is that what we want?
if (this.coordinator != null) && this.coordinator.isSnapshotInProgress()
in that case it will return false if coordinator is null

suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorTask.java
vaibhav-yb: Actually, coordinator being null means that the task is still initialisation phase and if we are saying that the connector has only two phases, i.e. snapshot and streaming then we can assume that the task is in streaming phase.

If coordinator is null, then we want the method to tell that we haven't moved on to the streaming phase so it is safe to return true.

If we change it to `if (this.coordinator != null) && this.coordinator.isSnapshotInProgress()` even this condition is okay but from a difference perspective. Consider that the coordinator is null, that is only in a situation when task is being initialised so `isTaskInSnapshotPhase()` will be `false`.

suranjan (approved):

PR #359, opened 2024-10-24
[DBZ][yugabyte/yugabyte-db#23903] CDCSDK: Disable transaction ordering with gRPC connector

vaibhav-yb:
This PR adds the following changes:
1. Disables the user to specify the `transaction.ordering` configuration - if a user specifies this, they will receive an error saying that transaction ordering is disabled and they should use logical replication instead.
2. Adds a configuration `TEST.override.transaction.ordering.deprecation` for internal uses in test until we finally remove the code.
3. Disables all the consistency related tests.


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1063,13 +1063,31 @@ public static AutoCreateMode parse(String value, String defaultValue) {
                     "'skip' to skip / ignore TRUNCATE events (default), " +
                     "'include' to handle and include TRUNCATE events");
 
+    public static final Field OVERRIDE_TRANSACTION_ORDERING_DEPRECATION = Field.create("override.transaction.ordering.deprecation")
+            .withDisplayName("Internal config to override and forcefully use transaction ordering")
+            .withImportance(Importance.LOW)
+            .withDefault(false)
+            .withType(Type.BOOLEAN);
+
     public static final Field TRANSACTION_ORDERING = Field.create("transaction.ordering")
```
Sumukh-Phalgaonkar, line 1072: Add a comment saying this field is deprecated.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1063,13 +1063,31 @@ public static AutoCreateMode parse(String value, String defaultValue) {
                     "'skip' to skip / ignore TRUNCATE events (default), " +
                     "'include' to handle and include TRUNCATE events");
 
+    public static final Field OVERRIDE_TRANSACTION_ORDERING_DEPRECATION = Field.create("override.transaction.ordering.deprecation")
```
Sumukh-Phalgaonkar, line 1066: Maybe change the name to `TEST.override.transaction.ordering.deprecation` and mention in description that this is test only config.

Sumukh-Phalgaonkar (approved):

PR #367, opened 2025-05-19
[yugabyte/yugabyte-db#27255] Modify YBExtractNewRecordState to introduce property "delete.to.tombstone"

vaibhav-yb:
## Problem

The current behaviour with the connector is that a tombstone record is published immediately after the delete record (when tombstones are enabled), this tombstone record has the same `OpId` as the delete record. A rough flow looks like:

```java
queue.enqueue(deleteRecord);

if (tombstonesEnabled) {
  queue.enqueue(tombstoneRecord);
}
```

Now there can be a situation where the connector crashes after adding the delete record to the queue and before it can add the tombstone record, it crashes or restarts - in that situation we might end up losing the tombstone record.

Since the tombstone record is essential to process delete records by some of the downstream connectors like the `io.confluent.connect.jdbc.JdbcSinkConnector` - these connectors will not be able to process deletes which would eventually be manifested as a data loss.

## Solution

To avoid getting into such a situation where we lose tombstone records, we have introduced a new configuration property here `delete.to.tombstone` which takes in a boolean value and it converts all the delete records to tombstone records. Now since we have the tombstone records already, we do not need the ones published by the connector so those records are dropped by this SMT.

> Another way to not get the original tombstone records is to disable tombstone publishing by the connector by setting `tombstones.on.delete=false` so that the connector doesn't produce any tombstone records itself.

### Testing

This PR also adds a test which can be run by using the command: 
```
mvn clean test -Dtest=YugabyteDBDatatypesTest#deletesShouldBeConvertedToTombstoneWhenTransformationEnabled
```

Closes yugabyte/yugabyte-db#27255.
Closes yugabyte/yugabyte-db#27048.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
@@ -19,6 +19,41 @@ import org.yb.util.Pair;
 import io.debezium.transforms.ExtractNewRecordState;
 import io.debezium.transforms.ExtractNewRecordStateConfigDefinition;
 
+/**
+ * This is a SMT used to flatten the records published by the YugabyteDB gRPC Connector. The connector
+ * publishes records which represent each operation that happen on a database - these records have a
+ * complex structure representing the details of the original database event.
+ *
+ * For example, a record representing an insert operation would have structure similar to the following:
+ * 
+ * <br/><br/>
+ * 
+ * <code>
+ *  "id": {"value": 123, "set": true},<br/>
+ *  "name": {"value": "John Doe", "set": true},<br/>
+ *  "email": {"value": "johndoe@email.com", "set": true}<br/>
+ * </code>
+ * 
+ * <br/>
+ * The complex structure published by the connector is not compatible with many downstream consumers
+ * and they expect a record in a simpler flattened format. This SMT extends the event flattening
+ * capabilities of the base {@link ExtractNewRecordState} SMT to facilitate event flattening.
+ *
+ * After flattening, the above record would look like:
+ * 
+ * <br/><br/>
+ * <code> {"id": 123, "name": "John Doe", "email": "johndoe@email.com"} </code>
+ * <br/><br/>
+ * 
+ * Now, the records published by the YugabyteDB gRPC Connector by default only contain the columns
+ * which were changed in the original database event. Similarly, there can be a case where a column
+ * can be explicitly set to null in the original database event. So in addition to the above
+ * flattening, this SMT also handles the case where it distinguishes between a column which was 
+ * explicitly set to null and a column which was not changed at all in the original database event
+ * and generates a flattened record accordingly.
+ *
+ * @author Vaibhav Kushwaha (vkushwaha@yugabyte.com)
+ */
 public class YBExtractNewRecordState<R extends ConnectRecord<R>> extends ExtractNewRecordState<R> {
     private static final Logger LOGGER = LoggerFactory.getLogger(YBExtractNewRecordState.class);
 
--- a/src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
@@ -50,6 +50,9 @@ public class YBExtractNewRecordState<R extends ConnectRecord<R>> extends Extract
         final R ret = super.apply(record);
         if (ret == null || (ret.value() != null && !(ret.value() instanceof Struct))) {
             // If ret == null, it means that the base SMT has dropped the record.
+            // The other condition is to check if the record is a valid envelope record
+            // (DML record) and not a metadata record like a transactional or heartbeat
+            // record, as the latter ones need to be passed through as is.
             LOGGER.trace("Returning the value as returned by the base transform");
             return ret;
         }
--- a/src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
@@ -29,9 +29,10 @@ public class YBExtractNewRecordState<R extends ConnectRecord<R>> extends Extract
 
     @Override
     public void configure(Map<String, ?> configs) {
-        // The config will be null when the user will not provide anything. We are falling back on
-        // Boolean.valueOf() in that case as it would return false when the provided argument is null.
-        convertDeleteToTombstone = Boolean.valueOf((String) configs.get(DELETE_TO_TOMBSTONE));
+        // Explicitly set default value to false when config is not provided.
+        String deleteToTombstoneConfig = (String) configs.get(DELETE_TO_TOMBSTONE);
+        convertDeleteToTombstone = deleteToTombstoneConfig != null ?
+                                    Boolean.parseBoolean(deleteToTombstoneConfig) : false;
 
         // Create a mutable copy of configs to allow modifications
         Map<String, Object> mutableConfigs = new java.util.HashMap<>(configs);
--- a/src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
@@ -6,6 +6,7 @@ import java.util.Objects;
 import org.apache.kafka.common.cache.Cache;
 import org.apache.kafka.common.cache.LRUCache;
 import org.apache.kafka.common.cache.SynchronizedCache;
+import org.apache.kafka.common.config.ConfigDef;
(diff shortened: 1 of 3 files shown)


src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
```diff
@@ -17,18 +17,51 @@
 import org.yb.util.Pair;
 
 import io.debezium.transforms.ExtractNewRecordState;
+import io.debezium.transforms.ExtractNewRecordStateConfigDefinition;
 
 public class YBExtractNewRecordState<R extends ConnectRecord<R>> extends ExtractNewRecordState<R> {
-    private Cache<Schema, Schema> schemaUpdateCache = new SynchronizedCache<>(new LRUCache<>(256));
```
vaibhav-yb, line 22: Removed this since it wasn't used anywhere in the code.

src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
```diff
@@ -41,6 +74,11 @@ public R apply(final R record) {
             updatedValueForValue = (Struct) val.getSecond();
         }
 
+        if (convertDeleteToTombstone && isDeleteRecord(record)) {
+            LOGGER.trace("Converting delete record to tombstone as per configuration");
+            return ret.newRecord(ret.topic(), ret.kafkaPartition(), updatedSchemaForKey, updatedValueForKey, null, null, record.timestamp());
```
Sumukh-Phalgaonkar, line 79: Don't we need to send the original delete record as well?

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
```diff
@@ -317,6 +318,59 @@ public void testRecordDeleteFieldWithYBExtractNewRecordState(boolean consistentS
...
+            transformation.configure(configs);
+
+            String dbStreamId = TestHelper.getNewDbStreamId("yugabyte", "t1", false, false);
+            Configuration.Builder configBuilder = TestHelper.getConfigBuilder("public.t1", dbStreamId);
+            startEngine(configBuilder);
+            final long rowsCount = 1;
+
+            awaitUntilConnectorIsReady();
+
+            // insert rows in the table t1 with values <some-pk, 'Vaibhav', 'Kushwaha', 30>
```
Sumukh-Phalgaonkar, line 338:
```suggestion
            // Insert rows in the table t1 with values <some-pk, 'Vaibhav', 'Kushwaha', 30>.
```

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
```diff
@@ -317,6 +318,59 @@ public void testRecordDeleteFieldWithYBExtractNewRecordState(boolean consistentS
...
+            insertRecords(rowsCount);
+
+            // delete rows in the table t1 where id is <some-pk>
```
Sumukh-Phalgaonkar, line 341:
```suggestion
            // Delete rows in the table t1 where id is <some-pk>.
```

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBDatatypesTest.java
```diff
@@ -317,6 +318,59 @@ public void testRecordDeleteFieldWithYBExtractNewRecordState(boolean consistentS
...
     }
 
+    @Test
+    public void deletesShouldBeConvertedToTombstoneWhenTransformationEnabled() throws Exception {
+        try (YBExtractNewRecordState<SourceRecord> transformation = new YBExtractNewRecordState<>()) {
+            TestHelper.dropAllSchemas();
+            TestHelper.executeDDL("yugabyte_create_tables.ddl");
+
+            Map<String, Object> configs = new HashMap<String, Object>();
+            configs.put(YBExtractNewRecordState.DELETE_TO_TOMBSTONE, "true");
```
Sumukh-Phalgaonkar, line 328: Should we add a test with `drop.tombstone` set to true?

src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
```diff
@@ -17,18 +17,51 @@
...
+        // Boolean.valueOf() in that case as it would return false when the provided argument is null.
+        convertDeleteToTombstone = Boolean.valueOf((String) configs.get(DELETE_TO_TOMBSTONE));
+
+        // Create a mutable copy of configs to allow modifications
+        Map<String, Object> mutableConfigs = new java.util.HashMap<>(configs);
+
+        if (convertDeleteToTombstone) {
+            // If we need to convert delete records to tombstones, we do not want
+            // the base SMT to drop delete records, set delete.handling.mode to 'none'.
+            mutableConfigs.put(ExtractNewRecordStateConfigDefinition.HANDLE_DELETES.name(), "none");
```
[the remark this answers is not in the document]
vaibhav-yb: As discussed offline, mentioning it here:

The base SMT is configured to drop delete records by default and the base SMT is applied first in the given method by using `super.apply(record)` - so if this newly introduced mode is configured, we do not want the base SMT to drop delete records as they will be converted to tombstone records by us. Hence, we provide the config to not drop delete records.

src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
```diff
@@ -17,18 +17,51 @@
...
+
+        super.configure(mutableConfigs);
+    }
+
     @Override
     public R apply(final R record) {
         final R ret = super.apply(record);
         if (ret == null || (ret.value() != null && !(ret.value() instanceof Struct))) {
+            // If ret == null, it means that the base SMT has dropped the record.
+            LOGGER.trace("Returning the value as returned by the base transform");
```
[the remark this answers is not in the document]
vaibhav-yb: The purpose of keeping it trace is that we will not be getting any new information by running it on `debug` level.

Think from the perspective of our test environments, at times we do specify a `LOG_LEVEL=debug` which automatically gets applied to the complete environment. There are other places in the connector where we print the records when log level is debug, now if we keep this as debug itself, we will end up having log spew without any real good information. That's why I have kept it to trace to only enable it in this class when absolutely needed.

src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
```diff
@@ -17,18 +17,51 @@
...
             return ret;
         }
 
+        // If the config drop.tombstone is set to true, they will be dropped by the base SMT
+        // which is handled in the previous block itself, but if it is set to false and our
+        // config delete.to.tombstone is set to true, we need to drop it here.
+        if (convertDeleteToTombstone && isTombstoneRecord(record)) {
+            LOGGER.trace("Dropping tombstone record as per configuration");
```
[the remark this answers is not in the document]
vaibhav-yb: Replied to above comment.

src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
vaibhav-yb: No, we do not need to send the original record. This SMT is used where we need record flattening and the purpose is to convert all delete records to tombstone records so that the SMT can finally consume them.

Sumukh-Phalgaonkar (approved):

src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
```diff
@@ -17,18 +17,55 @@
 import org.yb.util.Pair;
 
 import io.debezium.transforms.ExtractNewRecordState;
+import io.debezium.transforms.ExtractNewRecordStateConfigDefinition;
```
suranjan, line 20: Can we describe the behavioor in top level class comment

suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/transforms/YBExtractNewRecordState.java
```diff
@@ -17,18 +17,90 @@
...
+ * 
+ * Now, the records published by the YugabyteDB gRPC Connector by default only contain the columns
+ * which were changed in the original database event. Similarly, there can be a case where a column
+ * can be explicitly set to null in the original database event. So in addition to the above
+ * flattening, this SMT also handles the case where it distinguishes between a column which was 
+ * explicitly set to null and a column which was not changed at all in the original database event
+ * and generates a flattened record accordingly.
+ *
+ * @author Vaibhav Kushwaha (vkushwaha@yugabyte.com)
+ */
```
suranjan, line 56: where is the description for DELETE_TO_TOMBSTONE?

suranjan (changes requested):

PR #402, opened 2026-03-12
[yugabyte/yugabyte-db#30391] Handle NPE in isBeforeImageEnabled when using Logical Replication

bakul-gupta:
Jira Link: [DB-20273](https://yugabyte.atlassian.net/browse/DB-20273)
## Description

While validating gRPC connector support for Logical Replication (using a replication slot + publication instead of a stream created via yb-admin), the connector crashes during initialization with a NullPointerException in YBClientUtils.isBeforeImageEnabled().
```
Error while trying to get before image status
java.lang.NullPointerException
  at io.debezium.connector.yugabytedb.YBClientUtils.isBeforeImageEnabled(YBClientUtils.java:347)
  at io.debezium.connector.yugabytedb.YugabyteDBgRPCConnector.taskConfigs(YugabyteDBgRPCConnector.java:167)
```
## Solution

When using a replication slot + publication instead of a yb-admin stream, the CDC stream options may not contain a `record_type` key, causing `cdcStreamInfo.getOptions().get("record_type")` to return null and the subsequent .equals() call to throw an NPE. Added a null check in `YBClientUtils.isBeforeImageEnabled()` — if record_type is missing, log a warning and return true (before image enabled).

## Test Plan 
```
mvn test -Dtest=YugabyteDBPublicationReplicationTest
```

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
@@ -348,7 +348,8 @@ public class YBClientUtils {
     if (recordType == null) {
       // If the record_type option is not found, it means that the stream is configured with a slot + publication path.
       // In this case, we return true to indicate that before image is enabled.
-      LOGGER.warn("record_type option not found in stream options.");
+      LOGGER.info("record_type option not found in stream options, the stream is likely configured"
+          + " via slot + publication path; returning before_image as enabled (true)");
       return true;
     }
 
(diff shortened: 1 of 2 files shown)


Sumukh-Phalgaonkar (changes requested):

suranjan (changes requested):

suranjan:
Please add a test where 3 tables each with different replica identity is part of the same stream and all of them work with expected record format.

suranjan (approved):

PR #404, opened 2026-04-08
[yugabyte/yugabyte-db#31778] Add validation for mutually exclusive Stream ID and Slot Name

bakul-gupta:
Jira Link: [DB-20880](https://yugabyte.atlassian.net/browse/DB-20880)
## Description

The YugabyteDB connector supports two mutually exclusive CDC paths: a gRPC stream created via `yb-admin` (configured via `database.streamid`) and a logical replication slot + publication (configured via `slot.name`). When a user accidentally sets both `database.streamid` and a non-default `slot.name`, the connector previously did not detect the conflict at submit time, leading to ambiguous runtime behavior.

## Solution

Added connector-level validation that rejects configurations where both `database.streamid` and a non-default `slot.name` are set.

- Added `hasStreamIdAndSlotNameConflict` and `buildStreamIdAndSlotNameConflictMessage` helpers (private static) plus `assertStreamIdAndSlotNameMutuallyExclusive` (with an overload that accepts an additional message) in `YugabyteDBConnectorConfig`.
- Wired validation into `YugabyteDBgRPCConnector#validateConnection` so the error is surfaced on the `database.streamid` field at config-submit time, and into the connector start path so it runs once at the connector level rather than per task.
- Error message names both values, e.g. `Cannot set both stream id: <streamId> and a non-default slot name: <slotName>. Use slot.name (and publication.name) without a stream ID, or use a stream ID and leave slot.name at its default (<default>).`

## Test Plan
```
mvn test -Dtest=YugabyteDBConnectorConfigTest
mvn test -Dtest=YugabyteDBgRPCConnectorTest
```

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
@@ -626,7 +626,6 @@ public class YugabyteDBConnectorConfig extends RelationalDatabaseConnectorConfig
             .withGroup(Field.createGroupEntry(Field.Group.CONNECTION, 9))
             .withWidth(Width.MEDIUM)
             .withImportance(Importance.HIGH)
-            .withValidation(YugabyteDBConnectorConfig::validateStreamIdAndSlotNameMutuallyExclusive)
             .withDescription("ID of the Stream created in YugabyteDB");
 
     protected static final Field TASK_ID = Field.create("yugabytedb.task.id")
@@ -1589,17 +1588,6 @@ public class YugabyteDBConnectorConfig extends RelationalDatabaseConnectorConfig
             "Cannot set both %s and a non-default %s. "
                     + "Use slot.name (and publication.name) without a stream ID, "
                     + "or use a stream ID and leave slot.name at its default (%s).";
-
-    private static int validateStreamIdAndSlotNameMutuallyExclusive(Configuration config, Field field, Field.ValidationOutput problems) {
-        if (hasStreamIdAndSlotNameConflict(config)) {
-            problems.accept(field, config.getString(field),
-                    String.format(STREAM_ID_SLOT_CONFLICT_MSG,
-                            STREAM_ID.name(), SLOT_NAME.name(),
-                            ReplicationConnection.Builder.DEFAULT_SLOT_NAME));
-            return 1;
-        }
-        return 0;
-    }
 
     public static void assertStreamIdAndSlotNameMutuallyExclusive(Configuration config) {
         if (hasStreamIdAndSlotNameConflict(config)) {
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
@@ -626,6 +626,7 @@ public class YugabyteDBConnectorConfig extends RelationalDatabaseConnectorConfig
             .withGroup(Field.createGroupEntry(Field.Group.CONNECTION, 9))
             .withWidth(Width.MEDIUM)
             .withImportance(Importance.HIGH)
+            .withValidation(YugabyteDBConnectorConfig::validateStreamIdAndSlotNameMutuallyExclusive)
             .withDescription("ID of the Stream created in YugabyteDB");
 
     protected static final Field TASK_ID = Field.create("yugabytedb.task.id")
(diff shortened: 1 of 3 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1573,6 +1573,31 @@ public static ConfigDef configDef() {
...
     }
 
+    /**
+     * Returns {@code true} when both {@code database.streamid} and a non-default {@code slot.name}
+     * are configured. The two options represent mutually exclusive CDC paths.
+     */
+    public static boolean hasStreamIdAndSlotNameConflict(Configuration config) {
+        String streamId = config.getString(STREAM_ID);
+        String slotName = config.getString(SLOT_NAME);
+        return streamId != null && !streamId.isBlank()
```
shishir2001-yb: Please use `isEmpty()` to stay consistent with the rest of the codebase.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1573,6 +1573,31 @@ public static ConfigDef configDef() {
...
+            "Cannot set both %s and a non-default %s. "
+                    + "Use slot.name (and publication.name) without a stream ID, "
+                    + "or use a stream ID and leave slot.name at its default (%s).";
+
+    public static void assertStreamIdAndSlotNameMutuallyExclusive(Configuration config) {
+        if (hasStreamIdAndSlotNameConflict(config)) {
+            throw new DebeziumException(
+                    String.format(STREAM_ID_SLOT_CONFLICT_MSG,
+                            STREAM_ID.name(), SLOT_NAME.name(),
+                            ReplicationConnection.Builder.DEFAULT_SLOT_NAME));
```
shishir2001-yb: Take care of formatting

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1573,6 +1573,31 @@ public static ConfigDef configDef() {
...
+     */
+    public static boolean hasStreamIdAndSlotNameConflict(Configuration config) {
+        String streamId = config.getString(STREAM_ID);
+        String slotName = config.getString(SLOT_NAME);
+        return streamId != null && !streamId.isBlank()
+                && slotName != null && !slotName.equals(ReplicationConnection.Builder.DEFAULT_SLOT_NAME);
+    }
+
+    private static final String STREAM_ID_SLOT_CONFLICT_MSG =
+            "Cannot set both %s and a non-default %s. "
```
shishir2001-yb: > both %s and a non-default %s

When the error in printed it will be something like "Cannot set both : 7983478932... and a non-default slot_name". Instead can you make it like " both streamid %s and a non-default slot name: %s"

streamid and slot name: can be replaced with the args parameter i.e "Cannot set both: stream id: 7983478932... and a non-default slotname: slot_name"

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBgRPCConnector.java
```diff
@@ -262,6 +264,21 @@ protected void validateConnection(Map<String, ConfigValue> configValues, Configu
             return;
         }
 
+        if (YugabyteDBConnectorConfig.hasStreamIdAndSlotNameConflict(config)) {
+            String msg = String.format(
+                    "Cannot set both %s and a non-default %s. "
+                        + "Use slot.name (and publication.name) without a stream ID, "
+                        + "or use a stream ID and leave slot.name at its default. "
+                        + "Note: Publication is not supported with stream ID.",
```
shishir2001-yb: Why can't you use `assertStreamIdAndSlotNameMutuallyExclusive`? If you want to pass any additional message then add a parameter in the assertStreamIdAndSlotNameMutuallyExclusive function

shishir2001-yb (changes requested):
Please change the diff title, description, test plan in this format https://github.com/yugabyte/debezium-connector-yugabytedb/pull/402

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBgRPCConnector.java
bakul-gupta: Done — refactored to reuse` assertStreamIdAndSlotNameMutuallyExclusive`. Added an overload that takes an additionalMessage parameter, and the connector now calls it inside a try/catch (DebeziumException) to forward the message to `streamIdConfig.addErrorMessage(...)` (since validateConnection needs to surface errors per-field rather than throw). The conflict message is now built in one place via `buildStreamIdAndSlotNameConflictMessage`.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
bakul-gupta: Done.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
bakul-gupta: Done.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1573,6 +1573,44 @@ public static ConfigDef configDef() {
         return CONFIG_DEFINITION.configDef();
     }
 
+    /**
+     * Returns {@code true} when both {@code database.streamid} and a non-default {@code slot.name}
+     * are configured. The two options represent mutually exclusive CDC paths.
+     */
+    public static boolean hasStreamIdAndSlotNameConflict(Configuration config) {
```
shishir2001-yb: Should be private static

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -1573,6 +1573,44 @@ public static ConfigDef configDef() {
...
+        return streamId != null && !streamId.isEmpty()
+                && slotName != null && !slotName.equals(ReplicationConnection.Builder.DEFAULT_SLOT_NAME);
+    }
+
+    private static final String STREAM_ID_SLOT_CONFLICT_MSG =
+            "Cannot set both stream id: %s and a non-default slot name: %s. "
+            + "Use slot.name (and publication.name) without a stream ID, "
+            + "or use a stream ID and leave slot.name at its default (%s).";
+
+    public static String buildStreamIdAndSlotNameConflictMessage(Configuration config, String additionalMessage) {
```
shishir2001-yb: Should be private static

shishir2001-yb:
> Please change the diff title, description, test plan in this format https://github.com/yugabyte/debezium-connector-yugabytedb/pull/402

this is pending

shishir2001-yb (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
bakul-gupta: Done.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
bakul-gupta: Done.

bakul-gupta:
> Please change the diff title, description, test plan in this format #402

Done.

shishir2001-yb (approved):

PR #405, opened 2026-04-14
Add optimizations to connector bootstrap steps

jtanza:
## Problem
Connector validation in validateTServerConnection() makes a large number of redundant RPCs and creates unnecessary YBClient connections. On CDC streams with many tables, this causes validation to be significantly slower than necessary.

## Changes
 - `fetchTableList()` : New overload accepts a pre-fetched GetDBStreamInfoResponse for stream membership checks. The caller in validateTServerConnection() already has this response; previously it was re-fetched per table in the loop. Old signature preserved as a delegate.
 - `isBeforeImageEnabled() / isExplicitCheckpointingEnabled()`:  New overloads accept a pre-fetched CDCStreamInfo. taskConfigs() now calls getStreamInfo() once and reuses the result instead of fetching it twice.
 - `isYSQLStream()`: Results cached in a ConcurrentHashMap keyed by stream ID. The YugabyteDBConnectorConfig constructor calls this twice (once in super(), once for the field); the second call is now a cache hit.
 - `getTabletListToPollForCDCWithRetry()`:  New overload accepts an existing YBClient so the caller can reuse an open connection instead of creating a new one (with TLShandshake) per table. Used in validateTServerConnection(). The original signature used by streaming and snapshot sources is unchanged.

## Impact

 For a connector on a stream with T matched tables and N total tables in the stream:

   | | Before | After |
   |---|--------|-------|
   | `getDBStreamInfo` RPCs | 2 + N | 2 |
   | `getStreamInfo` calls | 2 | 1 |
   | `isYSQLStream` RPCs | 4 | 2 |
   | YBClient connections created | 4 + T | 2 |

 All changes are backward-compatible — new overloads only, no changes to existing method behavior.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
@@ -7,6 +7,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
@@ -41,6 +42,12 @@ import org.yb.master.MasterTypes.NamespaceIdentifierPB;
 public class YBClientUtils {
   private final static Logger LOGGER = LoggerFactory.getLogger(YBClientUtils.class);
 
+  /**
+   * Cache for isYSQLStream results keyed by stream ID. Since a stream's query language
+   * never changes, this is safe to cache indefinitely.
+   */
+  private static final ConcurrentHashMap<String, Boolean> isYSQLCache = new ConcurrentHashMap<>();
+
   public static boolean isTableIncludedInStreamId(GetDBStreamInfoResponse resp, String tableId) {
     for (MasterReplicationOuterClass.GetCDCDBStreamInfoResponsePB.TableInfo tableInfo : resp.getTableInfoList()) {
         if (Objects.equals(tableId, tableInfo.getTableId().toStringUtf8())) {
@@ -53,15 +60,35 @@ public class YBClientUtils {
   }
 
   /**
-   * Get the list of all the table UUIDs to be included for streaming
+   * Get the list of all the table UUIDs to be included for streaming.
+   * This overload fetches the DB stream info internally (one RPC per table — slow for large streams).
+   * Prefer the overload that accepts a pre-fetched {@link GetDBStreamInfoResponse}.
    * @param ybClient the {@link YBClient} instance
    * @param connectorConfig connector configuration for the connector
    * @return a Set of the tableIDs
    */
   public static Set<String> fetchTableList(YBClient ybClient,
                                            YugabyteDBConnectorConfig connectorConfig) {
+    try {
+      GetDBStreamInfoResponse streamInfoResponse = ybClient.getDBStreamInfo(connectorConfig.streamId());
+      return fetchTableList(ybClient, connectorConfig, streamInfoResponse);
+    } catch (Exception e) {
+      throw new DebeziumException(e);
+    }
+  }
+
+  /**
+   * Get the list of all the table UUIDs to be included for streaming, using a pre-fetched
+   * {@link GetDBStreamInfoResponse} to avoid redundant per-table RPCs.
+   * @param ybClient the {@link YBClient} instance
+   * @param connectorConfig connector configuration for the connector
+   * @param streamInfoResponse pre-fetched DB stream info response (used for stream membership checks)
+   * @return a Set of the tableIDs
+   */
+  public static Set<String> fetchTableList(YBClient ybClient,
+                                           YugabyteDBConnectorConfig connectorConfig,
+                                           GetDBStreamInfoResponse streamInfoResponse) {
     LOGGER.info("Fetching all the tables from the source");
-    String dbName = connectorConfig.getJdbcConfig().getDatabase();
     
     Set<String> tableIds = new HashSet<>();
       try {
@@ -88,23 +115,13 @@ public class YBClientUtils {
                               tableInfo.getName()));
                       continue;
                   }
-                  
-                  // Filter out the tables that are not in the database specified in the connector 
-                  // configuration
-                  // TODO(#29369): Filter out tables that don't belong to the database specified
-                  // in the connector config. This filtering should be done in YBClient itself.
-                  if (dbName != null
-                          && !dbName.equalsIgnoreCase(tableInfo.getNamespace().getName())) {
-                      continue;
-                  }
+
                   fqlTableName = tableInfo.getNamespace().getName() + "."
                                   + tableInfo.getPgschemaName() + "."
                                   + tableInfo.getName();
+                  tableId = YugabyteDBSchema.parseWithSchema(fqlTableName,
+                              tableInfo.getPgschemaName());
 
-                  tableId = YugabyteDBSchema.createTableIdWithCatalog(
-                    tableInfo.getNamespace().getName(),
-                    tableInfo.getPgschemaName(),
-                    tableInfo.getName());
               }
               else {
                   // Since there is no concept of schema in CQL we will be using namespaceName.tableName 
@@ -112,14 +129,11 @@ public class YBClientUtils {
                                   + tableInfo.getName();
                   tableId = YugabyteDBSchema.parseWithKeyspace(fqlTableName, tableInfo.getNamespace().getName());
               }
-              // Retrieve the list of tables in the stream ID,
-              GetDBStreamInfoResponse dbStreamInfoResponse = ybClient.getDBStreamInfo(
-                                                               connectorConfig.streamId());
 
               if (connectorConfig.getTableFilters().dataCollectionFilter().isIncluded(tableId)
                       && connectorConfig.databaseFilter().isIncluded(tableId)) {
                   // Throw an exception if the table in the include list is not a part of stream ID
-                  if (!isTableIncludedInStreamId(dbStreamInfoResponse, 
+                  if (!isTableIncludedInStreamId(streamInfoResponse, 
                                                  tableInfo.getId().toStringUtf8())) {
                       String warningMessageFormat = "The table %s is not a part of the "
                                                             + "stream ID %s. Ignoring the table.";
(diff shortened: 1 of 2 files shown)


shishir2001-yb:
@jtanza, could you share some numbers from your environment specifically a before vs. after comparison

jtanza:
> @jtanza, could you share some numbers from your environment specifically a before vs. after comparison

We have seen bootstrapping go from 7+ minutes in our staging cluster to seconds on the table filtering step:

```
kafka-connect-f7786556d-pzrjz kafka-connect 2026-08-10T18:37:24,008 WARN   ||  Filtering out the table {tableName} since it was not in the include list   [io.debezium.connector.yugabytedb.YBClientUtils] thread=pool-6-thread-176
kafka-connect-f7786556d-pzrjz kafka-connect 2026-08-10T18:37:24,031 WARN   ||  Filtering out the table {tableName}  since it was not in the include list   [io.debezium.connector.yugabytedb.YBClientUtils] thread=pool-6-thread-176
kafka-connect-f7786556d-pzrjz kafka-connect 2026-08-10T18:37:24,055 WARN   ||  Filtering out the table {tableName} since it was not in the include list   [io.debezium.connector.yugabytedb.YBClientUtils] thread=pool-6-thread-176
....
```

bdbene (approved):

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -88,38 +115,25 @@ public static Set<String> fetchTableList(YBClient ybClient,
...
-                  if (dbName != null
-                          && !dbName.equalsIgnoreCase(tableInfo.getNamespace().getName())) {
-                      continue;
-                  }
+
                   fqlTableName = tableInfo.getNamespace().getName() + "."
                                   + tableInfo.getPgschemaName() + "."
                                   + tableInfo.getName();
+                  tableId = YugabyteDBSchema.parseWithSchema(fqlTableName,
+                              tableInfo.getPgschemaName());
```
shishir2001-yb, line 123: This reverts 2fad9e6 ([#388](https://github.com/yugabyte/debezium-connector-yugabytedb/issues/388)), which built on deb3492 ([#386](https://github.com/yugabyte/debezium-connector-yugabytedb/issues/386), "Handle Ysql tables with space in the name"). Building a dotted string and re-parsing it is lossy.
```

table name | createTableIdWithCatalog(db, schema, name) | parseWithSchema(fql, schema)

normal_tbl | yugabyte.public.normal_tbl | same 
table with space | yugabyte.public.table with space | throws IllegalArgumentException: Unexpected input: w
tbl.with.dots | yugabyte.public.tbl.with.dots | silently truncated to yugabyte.public.tbl


```

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -88,38 +115,25 @@ public static Set<String> fetchTableList(YBClient ybClient,
...
                   }
-                  
-                  // Filter out the tables that are not in the database specified in the connector 
-                  // configuration
-                  // TODO(#29369): Filter out tables that don't belong to the database specified
-                  // in the connector config. This filtering should be done in YBClient itself.
-                  if (dbName != null
-                          && !dbName.equalsIgnoreCase(tableInfo.getNamespace().getName())) {
-                      continue;
-                  }
```
shishir2001-yb, line 99: Why do we need to remove it?

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -327,57 +341,72 @@ public static CDCStreamInfo getStreamInfo(YugabyteDBConnectorConfig connectorCon
...
-    }
-
-    return recordType.equals(CDCRecordType.ALL.name())
-           || recordType.equals(CDCRecordType.MODIFIED_COLUMNS_OLD_AND_NEW_IMAGES.name())
-           || recordType.equals(CDCRecordType.PG_FULL.name())
-           || recordType.equals(CDCRecordType.PG_CHANGE_OLD_NEW.name());
+    return (cdcStreamInfo.getOptions().get("record_type").equals(CDCRecordType.ALL.name())
+           || cdcStreamInfo.getOptions().get("record_type").equals(CDCRecordType.MODIFIED_COLUMNS_OLD_AND_NEW_IMAGES.name()))
+           || (cdcStreamInfo.getOptions().get("record_type").equals(CDCRecordType.PG_FULL.name())
+           || cdcStreamInfo.getOptions().get("record_type").equals(CDCRecordType.PG_CHANGE_OLD_NEW.name()));
```
shishir2001-yb, line 371: Exact inverse of b6c4a7a ([#402](https://github.com/yugabyte/debezium-connector-yugabytedb/issues/402), [yugabyte/yugabyte-db#30391](https://github.com/yugabyte/yugabyte-db/issues/30391)). The recordType == null guard is gone, so a stream with no record_type option NPEs on .get("record_type").equals(...): precisely the slot+publication path #402 fixed.

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -327,57 +341,72 @@ public static CDCStreamInfo getStreamInfo(YugabyteDBConnectorConfig connectorCon
...
 
       return cdcStreamInfo.getOptions().get("checkpoint_type")
               .equals(CdcService.CDCCheckpointType.EXPLICIT.name());
   }
 
   public static Boolean isYSQLStream(Configuration configuration) {
+    final String streamId = configuration.getString(YugabyteDBConnectorConfig.STREAM_ID);
+
+    // Return cached result if available — a stream's query language never changes.
+    Boolean cached = isYSQLCache.get(streamId);
```
shishir2001-yb, line 403: ConcurrentHashMap.get(null) throws NPE. shouldUsePublication() treats a null/empty stream.id as a legitimate config shape, so guard it, today that path gives a bare NPE with no log line, where previously you at least got Could not get Stream info for null.

shishir2001-yb, line 403: Use computeIfAbsent rather than get-then-put, otherwise two tasks starting concurrently both make the RPC.

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -53,15 +60,35 @@ public static boolean isTableIncludedInStreamId(GetDBStreamInfoResponse resp, St
   }
 
   /**
-   * Get the list of all the table UUIDs to be included for streaming
+   * Get the list of all the table UUIDs to be included for streaming.
+   * This overload fetches the DB stream info internally (one RPC per table — slow for large streams).
```
shishir2001-yb, line 64: This is no longer correct after your change?

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -460,6 +494,60 @@ public static GetTabletListToPollForCDCResponse getTabletListToPollForCDCWithRet
...
+   * @param ybClient existing client to reuse (caller manages lifecycle)
+   * @param table the {@link YBTable} instance of the table
+   * @param tableId the UUID of the table for which we need the tablets to poll for
+   * @param connectorConfig the configs used by the connector
+   * @return an RPC response containing the list of tablets to poll for
+   * @throws Exception when there are error after trying {@link YugabyteDBConnectorConfig#maxConnectorRetries()} times
+   */
+  public static GetTabletListToPollForCDCResponse getTabletListToPollForCDCWithRetry(
+      YBClient ybClient, YBTable table, String tableId,
+      YugabyteDBConnectorConfig connectorConfig) throws Exception {
```
shishir2001-yb, line 509: the old version took a fresh client per retry attempt, which recovers from a dead connection or master leader change, the new one retries on the same possibly-broken client. Probably fine since YBClient is designed to be long-lived? but was the retry path exercised under master failover? If per-attempt-fresh is intentional for the streaming sources, comment why.

shishir2001-yb (changes requested):
Stepping back from the inline comments: YBClientUtils.java in this PR is a very old version of the file, it predates #386, [#388](https://github.com/yugabyte/debezium-connector-yugabytedb/issues/388) and [#402](https://github.com/yugabyte/debezium-connector-yugabytedb/issues/402), and the three issues I flagged inline are those three fixes being silently reverted (they're exactly the only commits that touched this file since then). Rather than fixing the comments piecemeal, please re-apply your four optimization hunks onto a fresh checkout of current main. YugabyteDBgRPCConnector.java is unaffected. The optimizations themselves are all correct.

jtanza:
@shishir2001-yb re-applied the changes off of main. should be good to review now, thanks!

shishir2001-yb:
@jtanza, I am on PTO from tomorrow, i will pick it up early next week?

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -410,7 +477,15 @@ public static Boolean isYSQLStream(Configuration configuration) {
   }
 
   /**
-   * Call getTabletListToPollForCDC rpc with retries
+   * Call getTabletListToPollForCDC rpc with retries, creating a fresh {@link YBClient} for each
```
shishir2001-yb, line 480: The justification for the per-attempt client makes sense for retries, but it fires on attempt 1 too. In getChanges2 the caller already holds a healthy syncClient (line 351) and uses it for openTableByUUID one line earlier, so for T tables you pay T client constructions + TLS handshakes before any failure occurs. Fresh-client only helps after a failure. Worth using the caller's client for the first attempt and falling back to a fresh one on retry, which gets both properties

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -53,13 +60,43 @@ public static boolean isTableIncludedInStreamId(GetDBStreamInfoResponse resp, St
...
+   * This overload fetches the DB stream info once, up front, and then delegates to
+   * {@link #fetchTableList(YBClient, YugabyteDBConnectorConfig, GetDBStreamInfoResponse)}.
+   * Prefer that overload directly when the caller already holds a {@link GetDBStreamInfoResponse},
+   * to avoid the extra RPC.
    * @param ybClient the {@link YBClient} instance
    * @param connectorConfig connector configuration for the connector
    * @return a Set of the tableIDs
    */
   public static Set<String> fetchTableList(YBClient ybClient,
                                            YugabyteDBConnectorConfig connectorConfig) {
```
shishir2001-yb, line 74: Worth adding a test in some existing file if not already present?

shishir2001-yb (changes requested):
Overall the PR looks good, just some minor comments

Regarding diff summary:

Please cross check if numbers are correct here. 

   | | Before | After |
   |---|--------|-------|
   | `getDBStreamInfo` RPCs | 2 + N | 2 |
   | `getStreamInfo` calls | 2 | 1 |
   | `isYSQLStream` RPCs | 4 | 2 |
   | YBClient connections created | 4 + T | 2 |

shishir2001-yb (approved):

shishir2001-yb:
Thanks for your contribution, @jtanza . Merging the PR ✅

PR #410, opened 2026-07-21
[DBZ] Restore deprecated cdc.poll.interval.ms as an override for active/idle poll intervals

suranjan:
## Summary

PR #392 removed `cdc.poll.interval.ms` when it was split into `cdc.poll.interval.active.ms` and `cdc.poll.interval.idle.ms`. Existing configs that tuned the old property were silently ignored and fell back to the new defaults (10 ms active / 500 ms idle).

This change restores `cdc.poll.interval.ms` as a deprecated property:

- If set, its value is used for **both** the active and idle poll intervals.
- Any user-provided values of `cdc.poll.interval.active.ms` and `cdc.poll.interval.idle.ms` are ignored while it is set.
- A deprecation warning is logged (with a more specific message when the new properties are also set and being ignored).
- If not set, behavior is unchanged: the active/idle properties (or their defaults) apply.

The resolution logic lives in a package-private static method `YugabyteDBConnectorConfig.resolveCdcPollIntervals(Configuration)` so it can be unit-tested without a live cluster.

## Testing

- `mvn compile` passes locally.
- Added `YugabyteDBCdcPollIntervalTest` covering: defaults, active/idle only, legacy only, and legacy overriding active/idle. To be verified on devserver3.


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -687,6 +687,14 @@ public JdbcConfiguration getJdbcConfig() {
             .withImportance(Importance.LOW)
             .withDefault(DEFAULT_SOCKET_READ_TIMEOUT_MS);
```
shishir2001-yb, line 689: Add `@Deprecated`

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBCdcPollIntervalTest.java
```diff
@@ -0,0 +1,65 @@
+package io.debezium.connector.yugabytedb;
```
shishir2001-yb: Don't add a whole new test file just for this, please adjust it in some existing test.

shishir2001-yb (approved):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
suranjan: Done — added `@Deprecated` on `CDC_POLL_INTERVAL_MS` in 2a2339f.

---
_automated · Claude Code (Fable 5)_

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBCdcPollIntervalTest.java
suranjan: Done — removed the standalone test file and moved the four tests into `YugabyteDBConfigTest` in 2a2339f.

---
_automated · Claude Code (Fable 5)_

shishir2001-yb (approved):

PR #411, opened 2026-07-22
[DBZ][yugabyte/yugabyte-db#32752] Add per-tablet heartbeat support to the gRPC connector

shishir2001-yb:
## Problem

The gRPC connector never emits heartbeats. `YugabyteDBEventDispatcher` builds a heartbeat instance from the factory, but nothing ever fires it (there is a leftover `// TODO: Add heartbeat event processing here` in `dispatchDataChangeEvent`), so `heartbeat.interval.ms` has no effect on the gRPC path.

Because a single task polls many tablets, a tablet whose recent changes never reach Kafka, all dropped by `skipped.operations` or an SMT, or simply idle relative to the stream, stops advancing its explicit checkpoint. The server then retains that tablet's WAL and intents, which grow until retention limits are hit and can eventually fail the task. The worst shapes are a tablet parked on the split wait-list and a tablet whose `GetChanges` keeps failing: both stop being reached by the polling loop, so they go silent exactly when their checkpoint most needs to keep moving. The same applies during the initial snapshot: a long snapshot whose records are filtered never acknowledges anything, so a crash restarts the scan from the beginning.

## Solution

Emit a heartbeat per tablet, paced by a per-tablet timer, at the start of each tablet's turn in the polling loop, in both streaming sources and through the snapshot. Every tablet keeps advancing regardless of activity, and each heartbeat carries only its own tablet's checkpoint.

- `YugabyteDBStreamingChangeEventSource`
  - Added `maybeEmitTabletHeartbeat(part, offsetContext)`, called at the top of each per-tablet pass, before the split wait-list skip and before `GetChanges`. A tablet waiting on a split callback or failing its calls is never reached later in the pass, so beating first keeps parked and erroring tablets advancing. Pacing is per tablet via a `tabletToLastHeartbeatMs` map; there is no "is the tablet behind" condition, so idle, filtered, and active tablets all advance uniformly.
  - Added `heartbeatIntervalMs`, derived from `connectorConfig.getHeartbeatInterval()`.
- `YugabyteDBConsistentStreamingSource`
  - Calls the same helper at the top of its own per-tablet pass. The consistent source overrides the polling loop, so heartbeats had never fired under `transaction.ordering`.
- `YugabyteDBSnapshotChangeEventSource`
  - Added `maybeEmitSnapshotHeartbeat(part, previousOffset)`, per-tablet paced beats through the snapshot. Acked beats feed the explicit checkpoint that the next snapshot `GetChanges` ships, so the snapshot paging key keeps advancing in `cdc_state` and a mid-snapshot crash resumes from the last persisted key instead of rescanning the tablet. This matters most when every snapshot record is filtered, where no data record ever acknowledges anything.
  - Kept the completion burst: one beat per tablet when all tablets finish, flushing each tablet's snapshot-final checkpoint at the boundary.
  - The dispatcher field is now `YugabyteDBEventDispatcher<TableId>` (the constructor already receives that type), making the offset-map overload reachable.
- `YugabyteDBEventDispatcher`
  - Added `dispatchHeartbeatEvent(YBPartition, Map offset)`, which calls `Heartbeat.forcedBeat` keyed to the tablet's source partition. The base `dispatchHeartbeatEvent` relies on `HeartbeatImpl`'s single task-wide timer, which would let only one tablet beat per interval; `forcedBeat` plus the per-tablet timer above gives each tablet its own cadence. Heartbeats share the record queue, so a beat can never overtake a data record and confirm past it.
  - Added a `close()` override that closes the dispatcher's heartbeat instance, releasing its resources (including the JDBC connection when a heartbeat action query is configured).
- `YugabyteDBOffsetContext`
  - Added `getOffset(YBPartition)`, an offset naming only the given tablet. Every heartbeat (streaming, snapshot pacing, completion burst) carries just its own tablet's checkpoint, so a beat for one tablet can never confirm a sibling past records that were not acknowledged. The method is identical to the one introduced by #413, so the two PRs merge cleanly in either order.

Under explicit checkpointing, an acked heartbeat advances that tablet's checkpoint through the existing `commit()` -> `commitOffset()` -> `GetChanges` -> `cdc_state` path, which is what releases the retained WAL.

Heartbeats stay opt-in: `heartbeat.interval.ms` keeps the upstream default of `0` (disabled), and a test now pins that default. Set a positive value to enable. Note that the heartbeat topic must exist or be creatable; on clusters with topic auto-creation disabled, pre-create `__debezium-heartbeat.<topic.prefix>` before enabling.

## Testing

All docker-based, in `YugabyteDBHeartbeatTest`. Every scenario test verifies the checkpoint on the server through `YBClient.getCheckpoint` (`cdc_state`), not only that heartbeat records appear.

- `heartbeatCarriesCheckpointPastFilteredUpdatesAndStreamingResumes`: single tablet, `heartbeat.interval.ms=10000`, `skipped.operations=u`. Inserts 10 rows (delivered), applies 10 updates (filtered), idles ~65s. Asserts at least 5 beats arrive during the idle window, the beat checkpoint is past the last delivered insert, the server checkpoint moved past its pre-idle value, and later inserts still stream through.
- `snapshotCompletionEmitsHeartbeatPerTablet`: 3-tablet table populated before the connector starts, so rows flow through the initial snapshot. The interval is set large enough that the streaming timer cannot fire, so any heartbeat can only be the completion burst. Asserts all 30 rows are delivered, exactly one beat per tablet, each carrying at least the last snapshot position, and a server checkpoint set for every tablet.
- `heartbeatsAdvanceATabletWhoseEveryRecordIsFiltered`: range table split at 5000, seeded on both sides before the stream exists (no snapshot), then only updates, all dropped by `skipped.operations=u`. Asserts zero data records were produced, beats flow for both tablets, and both tablets' server checkpoints advance past their baselines.
- `heartbeatsFlowWhenEverySnapshotRecordIsFiltered`: `snapshot.mode=initial` with `skipped.operations=r`, so every snapshot record is dropped. Asserts per-tablet beats, zero data records, and a server checkpoint set for every tablet.
- `heartbeatOffsetCarriesOnlyItsOwnTablet`: 3-tablet table; waits until all three tablets appear across the beats, then asserts every heartbeat's offset references exactly one tablet.
- `heartbeatsAreOffByDefault`: no `heartbeat.interval.ms` configured; streams rows, idles, asserts zero heartbeat records. Pins the opt-in default.
- Manual end-to-end on a real cluster: range-split table with 3 tablets, one deliberately idle, `heartbeat.interval.ms=10000`, Kafka Connect with snapshot plus streaming for 10 minutes. All three tablets, including the idle one, produced a heartbeat on the `__debezium-heartbeat` topic every ~10s. This run predates the per-tablet offset change, so its beats still carried the whole map; the single-tablet payload is covered by `heartbeatOffsetCarriesOnlyItsOwnTablet`.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -62,7 +62,7 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
     private final YugabyteDBSchema schema;
     private final SnapshotProgressListener snapshotProgressListener;
     private final YugabyteDBTaskContext taskContext;
-    private final EventDispatcher<YBPartition,TableId> dispatcher;
+    private final YugabyteDBEventDispatcher<TableId> dispatcher;
     protected final Clock clock;
     private final Snapshotter snapshotter;
     private final YugabyteDBConnection connection;
@@ -394,10 +394,8 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
     }
 
     /**
-     * Emit a heartbeat for the given tablet if the heartbeat interval has elapsed since its last one.
-     * A snapshot can run for a long time and a single heartbeat at completion would be lost if the
-     * task died before it was flushed, so tablets are paced through the snapshot the same way they
-     * are while streaming.
+     * Emit a heartbeat for the tablet if its interval has elapsed. A snapshot can run for a long time
+     * and a single beat at completion would be lost if the task died before it was flushed.
      */
     private void maybeEmitSnapshotHeartbeat(YBPartition part, YugabyteDBOffsetContext offsetContext) {
         if (!dispatcher.heartbeatsEnabled()) {
@@ -411,7 +409,7 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
         }
 
         try {
-            dispatcher.alwaysDispatchHeartbeatEvent(part, offsetContext);
+            dispatcher.dispatchHeartbeatEvent(part, offsetContext.getOffset(part));
             tabletToLastSnapshotHeartbeatMs.put(part.getId(), currentTimeMs);
         }
         catch (InterruptedException e) {
@@ -525,7 +523,7 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
                     if (dispatcher.heartbeatsEnabled()) {
                         for (Pair<String, String> tablet : tableToTabletForSnapshot) {
                             YBPartition hbPartition = new YBPartition(tablet.getKey(), tablet.getValue(), true);
-                            dispatcher.alwaysDispatchHeartbeatEvent(hbPartition, previousOffset);
+                            dispatcher.dispatchHeartbeatEvent(hbPartition, previousOffset.getOffset(hbPartition));
                         }
                     }
                     return SnapshotResult.completed(previousOffset);
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
@@ -82,6 +82,10 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
     protected List<HashPartition> partitionRanges;
 
     private boolean snapshotComplete = false;
+
+    // Per-tablet last heartbeat emit time (ms), to pace heartbeats through a long snapshot.
+    private final Map<String, Long> tabletToLastSnapshotHeartbeatMs = new HashMap<>();
+    private final long heartbeatIntervalMs;
     private Map<String, Long> lastGetChangesTime;
     private final String LAST_SNAPSHOT_RECORD_KEY = "LAST_SNAPSHOT_RECORD";
 
@@ -102,6 +106,7 @@ public class YugabyteDBSnapshotChangeEventSource extends AbstractSnapshotChangeE
         this.snapshotter = snapshotter;
         this.connection = connection;
         this.snapshotProgressListener = snapshotProgressListener;
+        this.heartbeatIntervalMs = connectorConfig.getHeartbeatInterval().toMillis();
 
         AsyncYBClient asyncClient = new AsyncYBClient.AsyncYBClientBuilder(connectorConfig.masterAddresses())
             .defaultAdminOperationTimeoutMs(connectorConfig.adminOperationTimeoutMs())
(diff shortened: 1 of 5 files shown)


suranjan (commented):
Do heartbeat only apply to streaming and not snapshot?

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBHeartbeatTest.java
```diff
@@ -0,0 +1,238 @@
...
+                }
+                int dot = key.lastIndexOf('.');
+                tablets.add(dot < 0 ? key : key.substring(dot + 1));
+            }
+        }
+        return tablets;
+    }
+
+    @Test
+    public void snapshotCompletionEmitsHeartbeatPerTablet() throws Exception {
```
suranjan, line 215: 1. Add a test where all snapshot records are filtered.

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBHeartbeatTest.java
```diff
@@ -0,0 +1,238 @@
...
+                }
+                catch (NumberFormatException ignored) {
+                }
+            }
+        }
+        return max;
+    }
+
+    @Test
+    public void heartbeatCarriesCheckpointPastFilteredUpdatesAndStreamingResumes() throws Exception {
```
suranjan, line 129: Add a test where all records of a single tablet gets filtered. Maybe use a partition table or range etc.

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBHeartbeatTest.java
```diff
@@ -0,0 +1,238 @@
...
+import java.util.stream.Collectors;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Verifies the per-tablet timer heartbeat: with updates filtered out (skipped.operations=u), the tablet
+ * keeps advancing its checkpoint, the heartbeats carry an offset past the last delivered insert, and
+ * streaming still resumes for later inserts.
+ */
+public class YugabyteDBHeartbeatTest extends YugabyteDBContainerTestBase {
```
suranjan, line 35: No where we are verifying the explicit checkpoint. Just sending the heartbeat is not enough, need to be validated that explicit checkpoint moves forward.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -487,6 +487,14 @@ protected SnapshotResult<YugabyteDBOffsetContext> doExecute(ChangeEventSourceCon
                 if (snapshotCompletedTablets.size() == tableToTabletForSnapshot.size()) {
                     LOGGER.info("Snapshot completed for all the tablets");
                     this.snapshotComplete = true;
+                    // Emit one heartbeat per tablet at snapshot completion so each tablet's checkpoint is
```
suranjan, line 521: We send it only once? What if there is a crash. The snapshot completed heartbeat should be sent periodically just like streaming one with last offset.

suranjan: What does other connector do here? Check with YB logical snapshot heartbeat for reference.

suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
```diff
@@ -169,6 +169,18 @@ public OpId snapshotDoneKeyLsn(YBPartition partition) {
         return result;
     }
 
+    public Map<String, ?> getOffset(YBPartition partition) {
```
suranjan, line 174: This is with offst diff?

suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
shishir2001-yb: This is also with offset diff(needed here too), will rebase before merging this

suranjan (approved):

PR #413, opened 2026-07-23
[DBZ][yugabyte/yugabyte-db#32753] Carry only the record's own tablet checkpoint in its offset

shishir2001-yb:
## Problem

A task polls many tablets, but `YugabyteDBOffsetContext.getOffset()` returns the whole per-task map, so every record emitted for any tablet carries a snapshot of every tablet's checkpoint. Kafka Connect stores that map under the producing tablet's source partition, and `commit()` -> `commitOffset()` then applies every key in it.

The consequence is that one tablet's acknowledgement confirms its siblings. If a sibling has records that were consumed but never reached Kafka, still queued, or dropped by an SMT after dispatch, its checkpoint is confirmed past them anyway. `cdc_state` moves beyond records Kafka never stored and a restart resumes after them, so the rows are lost silently. Under a broker outage that leaves some records unacknowledged while others keep flowing, this reproduces as committed rows missing from the sink with no error anywhere.

This shape dates back to EXPLICIT checkpointing (#179), where the Debezium 1.9.5 `BaseSourceTask` handed `commitOffset` exactly one record's offset map, so only a whole-map record could confirm all of a task's tablets in one callback. #348 removed that constraint by reading all per-partition committed offsets back from Kafka on every cycle, so the record shape is no longer load bearing.

The empty transaction COMMIT position advance, which per-tablet offsets would otherwise expose, is split into its own change; this PR assumes it lands first.

## Solution

Give each record an offset that names only its own tablet, and close a gap that the whole map was quietly covering.

- `YugabyteDBOffsetContext`
  - Added `getOffset(YBPartition)`, which filters `tabletSourceInfo` to the requested tablet, matching both the bare `tabletId` and the `tableId.tabletId` key form used for colocated tables. The no-arg `getOffset()` is unchanged for callers that legitimately need the whole map.
- `YugabyteDBEventDispatcher`, `YugabyteDBTransactionMonitor`, `LogicalDecodingMessageMonitor`
  - Build `SourceRecord`s with `getOffset(partition)` instead of `getOffset()`. These are the three places records are emitted: data changes, transaction metadata, and logical decoding messages. All of them are shared by the vanilla and the consistent streaming sources, so both paths are covered.
- `YugabyteDBConnectorTask`
  - Skip the commit callback when the task has no stored offset for any of its partitions yet. `ybOffset` is only populated from partitions that have a stored offset, so a task whose tablets have not produced anything reached `coordinator.commitOffset(null)`.
- `YugabyteDBgRPCConnector`
  - Return early from `stop()` when the table monitoring thread was never started, so a connector that failed validation reports its original failure instead of an NPE.

A tablet with no records of its own still advances: the empty-poll self advance added in #246 moves its checkpoint to the response checkpoint whenever everything it has consumed is already confirmed. That path is gated on `lastRecordCheckpoint <= explicitCheckpoint`, so unlike a cross-tablet acknowledgement it can never confirm past a record the connector took responsibility for.

Known limitation, unchanged by this PR: a tablet whose entire tail is dropped after dispatch, by `skipped.operations` or an SMT, has its consumed position moved with nothing produced, so its checkpoint freezes. Per-tablet heartbeats (#411) are the fix for that case.

## Testing

All docker-based, in `YugabyteDBPerTabletOffsetTest`.

- `recordOffsetCarriesOnlyItsOwnTablet`: 2-tablet table, 400 inserts spread wide enough that both tablets deliver. Asserts every data record's `sourceOffset()` names exactly one tablet, and that both tablets actually produced so the assertion is not vacuous.
- `activeTabletAdvancesCheckpointWithoutDraggingFilteredSibling`: range table split at 500 with `skipped.operations=u`. After a baseline where both tablets deliver, one tablet keeps receiving delivered inserts while the other receives only filtered updates. Asserts the active tablet's server checkpoint moves forward and the filtered sibling is not confirmed past its 40 filtered updates. The sibling is allowed a small gain because the self advance legitimately nudges it; a cross-tablet drag would show up as a gain of 40 or more.
- `everyTabletAdvancesItsOwnCheckpointOnServer`: 3-tablet table, 400 inserts, then reads `cdc_state` through `YBClient.getCheckpoint` for each tablet and requires every index to be past the stream's starting position. Verifies the checkpoint really moved on the server rather than only that records were emitted.
- `checkpointAdvancesForTabletThatProducedNoRecords`: range table split at 5000 with all inserts confined below the split point, so one tablet streams and the other is silent. Asserts exactly one tablet appears in the record offsets, and that both tablets' server checkpoints still advanced. This is the regression guard for the silent tablet, which no longer has a sibling to carry it.
- `recordOffsetNeverCarriesASiblingTabletUnderTransactionMetadata`: same setup with `provide.transaction.metadata=true`. Asserts every captured record, including the BEGIN and END records on the transaction topic, references at most one tablet. Transaction metadata is emitted through a different path than data records and had no coverage otherwise.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
@@ -169,6 +169,20 @@ public class YugabyteDBOffsetContext implements OffsetContext {
         return result;
     }
 
+    // Offset carrying only the given tablet's checkpoint, so a record never advances a sibling
+    // tablet.
+    public Map<String, ?> getOffset(YBPartition partition) {
+        Map<String, Object> result = new HashMap<>();
+        String tabletId = partition.getTabletId();
+        for (Map.Entry<String, SourceInfo> entry : this.tabletSourceInfo.entrySet()) {
+            String key = entry.getKey();
+            if (key.equals(tabletId) || key.endsWith("." + tabletId)) {
+                result.put(key, entry.getValue().lsn().toSerString());
+            }
+        }
+        return result;
+    }
+
     public Struct getSourceInfoForTablet(YBPartition partition) {
         return this.tabletSourceInfo.get(partition.getId()).struct();
     }
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
@@ -658,8 +658,15 @@ public class YugabyteDBStreamingChangeEventSource implements
                                             }
                                             if (message.getOperation() == Operation.COMMIT) {
                                                 LOGGER.debug("LSN in case of COMMIT is " + lsn);
-                                                offsetContext.updateRecordPosition(part, lsn, lastCompletelyProcessedLsn, message.getRawCommitTime(),
-                                                        String.valueOf(message.getTransactionId()), null, message.getRecordTime(), message.getXreplOriginId());
+                                                // A transactional block with no records for this tablet has nothing that can be
+                                                // produced and therefore nothing that can ever be acknowledged, so its position
+                                                // must not become the last seen record's checkpoint.
+                                                boolean emptyTransactionalBlock = recordsInTransactionalBlock.containsKey(part.getId())
+                                                        && recordsInTransactionalBlock.get(part.getId()) == 0;
+                                                if (!emptyTransactionalBlock) {
+                                                    offsetContext.updateRecordPosition(part, lsn, lastCompletelyProcessedLsn, message.getRawCommitTime(),
+                                                            String.valueOf(message.getTransactionId()), null, message.getRecordTime(), message.getXreplOriginId());
+                                                }
 
                                                 if (recordsInTransactionalBlock.containsKey(part.getId())) {
                                                     if (recordsInTransactionalBlock.get(part.getId()) == 0) {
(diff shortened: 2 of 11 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
```diff
@@ -169,6 +169,20 @@ public OpId snapshotDoneKeyLsn(YBPartition partition) {
         return result;
     }
 
+    // Offset carrying only the given tablet's checkpoint, so a record never advances a sibling
+    // tablet.
+    public Map<String, ?> getOffset(YBPartition partition) {
```
suranjan, line 174: Why not return YugabyteDBOffsetContext

suranjan (changes requested):

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -658,8 +658,11 @@ protected void getChanges2(ChangeEventSourceContext context,
                                             }
                                             if (message.getOperation() == Operation.COMMIT) {
                                                 LOGGER.debug("LSN in case of COMMIT is " + lsn);
-                                                offsetContext.updateRecordPosition(part, lsn, lastCompletelyProcessedLsn, message.getRawCommitTime(),
-                                                        String.valueOf(message.getTransactionId()), null, message.getRecordTime(), message.getXreplOriginId());
+                                                Integer recordsInBlock = recordsInTransactionalBlock.get(part.getId());
+                                                if (recordsInBlock == null || recordsInBlock != 0) {
+                                                    offsetContext.updateRecordPosition(part, lsn, lastCompletelyProcessedLsn, message.getRawCommitTime(),
```
shishir2001-yb: TODO: Maybe add a new diff for this particular change

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
shishir2001-yb: SourceRecord expects a Map not `YugabyteDBOffsetContext`. 
We need to type cast `offsetContext` because it is calling getOffset(arg) a method which is not present in upstream only introduced in `YugabyteDBOffsetContext`. 

new SourceRecord(
partition.getSourcePartition(),
 ((**YugabyteDBOffsetContext) offsetContext**).getOffset(partition),

Without typecasting we get the below compilation error

```
[ERROR] /tmp/castdemo_wt/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBEventDispatcher.java:[206,28] method getOffset in interface io.debezium.pipeline.spi.OffsetContext cannot be applied to given types;
  required: no arguments
  found:    io.debezium.connector.yugabytedb.YBPartition
  reason: actual and formal argument lists differ in length
[INFO] 1 error
```

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
shishir2001-yb: Done

suranjan (approved):

PR #420, opened 2026-08-19
Add connector name tag to JMX metrics

bdbene:
## Problem

The `server` tag on the task- and partition-scoped MBeans this connector registers is the Debezium logical name, i.e. the topic prefix. That is a property of the source database rather than of the connector, so every connector configured against the same database registers with the same `server` tag.

When more than one such connector runs in a Kafka Connect worker, this has two consequences:

1. **Metrics cannot be attributed to a connector.** `MilliSecondsBehindSource` and the event counters are only distinguishable down to the tablet in the `partition` tag, so lag reads as a property of the whole database. Recovering which connector a tablet belongs to means resolving table and tablet UUIDs out of band.

2. **Task-scoped MBeans collide.** `AbstractYugabyteDBTaskMetrics` registers at `server=<prefix>, task=<id>, context=<context>` with no `partition` tag, which is identical across connectors. Only the first connector to register succeeds; the rest exhaust the retry loop in `YugabyteDBMetrics#register` and log `Failed to register metrics MBean, metrics will not be available`. The observable symptom is one set of queue metrics per worker instead of one per connector.

## Change

Adds the name Kafka Connect assigned to the connector as a `connector` tag on the task- and partition-scoped MBeans.

- `YugabyteDBConnectorConfig#getConnectorMetricTags()` returns a `connector` tag read from Connect's `name` property, or an empty map when the configuration carries no `name` at all. A separate accessor is used because `getConnectorName()` is already overridden to return `Module.name()`, and `getLogicalName()` is the topic prefix.
- Each construction site appends it with `tags.putAll(connectorConfig.getConnectorMetricTags())`: the multi-partition task-level tag maps in `YugabyteDBMetrics`, and the partition tag maps in `YugabyteDBStreamingTaskMetrics` and `YugabyteDBSnapshotTaskMetrics`.

Mutating the map returned by `Collect.linkMapOf` is how Debezium core adds tags beyond the fixed set: `io.debezium.metrics.Metrics` has done exactly this for its `custom.metric.tags` since 2.6, and `Collect` offers no ordered-map factory past four pairs. Order matters because `metricName` serialises the map by iteration order.

**The tag is appended after the existing tags**, so any `ObjectName` pattern anchored on the leading tags keeps matching. `metricName(connectorType, connectorName, contextName)` is left untouched. `AbstractYugabyteDBTaskMetrics` always requests multi-partition mode, so the single-partition path is unaffected by this change.

## Effect on MBean names

`name` is set by Kafka Connect and propagated to every task, and `EmbeddedEngine.ENGINE_NAME` is the required field `name`, so the embedded engine and Debezium Server supply it too. The tag is therefore present in every deployment, and MBean names change accordingly — worth a release note. The empty-map path only applies to a `Configuration` built by hand, as in the unit tests below.

Anything matching MBean names by pattern has to account for the extra tag. `metrics.yml` in this repo gets two rules for the tagged beans, and the existing `debezium` rules have their trailing capture groups constrained from `[^>]+` to `[^,>]+` so that a tagged bean cannot fall through to a rule that silently absorbs the tag into the preceding label.

## Testing

- `mvn test-compile` passes.
- Three tests in `YugabyteDBConfigTest`. Two follow the shape of the existing `resolveCdcPollIntervals` tests against a static resolver taking a `Configuration`: one pins the resulting tag order, since exporter configurations depend on the tag being last, and one covers the absent-name case. The third starts the connector and asserts that every registered `debezium.yugabytedb:type=connector-metrics,*` MBean carries the tag, which covers the construction sites rather than the resolver.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
@@ -12,6 +12,7 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.time.Duration;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -56,6 +57,7 @@ import io.debezium.relational.RelationalTableFilters;
 import io.debezium.relational.TableId;
 import io.debezium.relational.Tables.TableFilter;
 import io.debezium.util.Clock;
+import io.debezium.util.Collect;
 import io.debezium.util.Metronome;
 import io.debezium.util.Strings;
 
@@ -616,6 +618,8 @@ public class YugabyteDBConnectorConfig extends RelationalDatabaseConnectorConfig
     public static final int DEFAULT_MBEAN_REGISTRATION_RETRIES = 12;
     public static final long DEFAULT_MBEAN_REGISTRATION_RETRY_DELAY_MS = 5_000;
     public static final long DEFAULT_LAST_CALLBACK_TIMEOUT_MS = 3 * 60 * 1000;
+
+    private static final String KAFKA_CONNECT_NAME = "name";
 
     @Override
     public JdbcConfiguration getJdbcConfig() {
@@ -2051,6 +2055,23 @@ public class YugabyteDBConnectorConfig extends RelationalDatabaseConnectorConfig
     @Override
     public String getConnectorName() {
         return Module.name();
+    }
+
+    /**
+     * Returns metric tags identifying this connector instance. {@link #getLogicalName()} is only the
+     * topic prefix, which every connector reading from the same database shares, whereas the name
+     * Kafka Connect assigned to the connector is unique within a worker.
+     *
+     * @return a single {@code connector} tag, or an empty map when the configuration carries no
+     *         {@code name} property
+     */
+    public Map<String, String> getConnectorMetricTags() {
+        return resolveConnectorMetricTags(getConfig());
+    }
+
+    static Map<String, String> resolveConnectorMetricTags(Configuration config) {
+        String connectorName = config.getString(KAFKA_CONNECT_NAME, "");
+        return connectorName.isEmpty() ? Collections.emptyMap() : Collect.linkMapOf("connector", connectorName);
     }
 
     public ConnectionFactory getConnectionFactory() {
--- a/src/main/java/io/debezium/connector/yugabytedb/metrics/YugabyteDBMetrics.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/metrics/YugabyteDBMetrics.java
@@ -62,10 +62,12 @@ public class YugabyteDBMetrics {
     registrationRetryDelay = Duration.ofMillis(connectorConfig.mbeanRegistrationRetryDelayMs());
 
     if (multiPartitionMode) {
-      this.name = metricName(connectorType, Collect.linkMapOf(
+      Map<String, String> tags = Collect.linkMapOf(
               "server", connectorName,
               "task", connectorConfig.getTaskId(),
-              "context", contextName));
+              "context", contextName);
+      tags.putAll(connectorConfig.getConnectorMetricTags());
+      this.name = metricName(connectorType, tags);
     } else {
       this.name = metricName(connectorType, connectorName, contextName);
     }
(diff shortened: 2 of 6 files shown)


shishir2001-yb:
@bdbene, can this be reviewed?

bdbene:
I have this running and working, but I rushed it out and I'll clean it up before review to improve quality.

bdbene:
@shishir2001-yb I cleaned it up, and took some inspiration from newer versions of Debezium for how to have more than 4 labels on a metric.

shishir2001-yb:
Added this to my queue

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -617,6 +619,10 @@ public static ConsistencyMode parse(String value) {
     public static final long DEFAULT_MBEAN_REGISTRATION_RETRY_DELAY_MS = 5_000;
     public static final long DEFAULT_LAST_CALLBACK_TIMEOUT_MS = 3 * 60 * 1000;
 
+    // The property Kafka Connect uses to pass a connector's name down to its tasks. Declared here to
+    // avoid a compile dependency on connect-runtime.
```
shishir2001-yb: We can remove this comment

src/test/java/io/debezium/connector/yugabytedb/YugabyteDBConfigTest.java
```diff
@@ -146,6 +148,28 @@ public void shouldIgnoreActiveAndIdlePollIntervalsWhenDeprecatedPropertyIsSet()
         assertEquals(300, intervals[1]);
     }
 
+    @Test
```
shishir2001-yb, line 158: `shouldTagMetricsWithTheConnectorNameAfterTheExistingTags` rebuilds the tag map itself, so if one of the four putAll call sites is dropped (or taskConfigs() stops copying name through), no test fails. Since the test framework already runs the embedded engine with name=test-connector, an end to end pin is cheap: in any test that starts a connector, query `ManagementFactory.getPlatformMBeanServer().queryNames(new ObjectName("debezium.yugabytedb:type=connector-metrics,*"), null)` and assert the names contain connector=test-connector with the tag last (getKeyPropertyListString() preserves order). Alternatively, a unit test that calls metricName() through a small test subclass of YugabyteDBMetrics would cover the real serialization path.

src/main/java/io/debezium/connector/yugabytedb/metrics/YugabyteDBMetrics.java
```diff
@@ -91,12 +93,14 @@ protected YugabyteDBMetrics(String taskId,
...
-                  connectorType, connectorName, taskId, contextName);
-      this.name = metricName(connectorType, Collect.linkMapOf(
+      LOGGER.info("Configuring a metric with connector type {} server {}, task ID {}, context {} and connector tags {}",
+                  connectorType, connectorName, taskId, contextName, connectorConfig.getConnectorMetricTags());
+      Map<String, String> tags = Collect.linkMapOf(
           "server", connectorName,
           "task", taskId,
-          "context", contextName));
+          "context", contextName);
+      tags.putAll(connectorConfig.getConnectorMetricTags());
```
shishir2001-yb: The INFO log calls connectorConfig.getConnectorMetricTags() and then putAll calls it again. Build the tags map first and log the finished map

shishir2001-yb (commented):
This repo ships `metrics.yml` in the Docker image (Dockerfile:25) as the JMX exporter config, and its two Debezium rules end in greedy captures: partition=([^>]+) and context=([^>]+). With the new tag appended, the partition label becomes `<tablet>, connector=<name>` and the task-level context label becomes `streaming, connector=<name>`: mangled labels for anyone using our image. Please update `metrics.yml` in this PR: add connector-aware rules above the existing ones and tighten the old rules' final groups to [^,>]+ as fallback for the no-name case:

```
- pattern: "debezium.([^:]+)<type=connector-metrics, server=([^,]+), task=([^,]+), context=([^,]+), partition=([^,>]+), connector=([^>]+)>([^:]+)"
  name: "debezium_metrics_$7"
  labels: { plugin: "$1", name: "$2", context: "$4", task: "$3", partition: "$5", connector: "$6" }
- pattern: "debezium.([^:]+)<type=connector-metrics, server=([^,]+), task=([^,]+), context=([^,]+), connector=([^>]+)>([^:]+)"
  name: "debezium_metrics_$6"
  labels: { plugin: "$1", name: "$2", context: "$4", task: "$3", connector: "$5" }
```

shishir2001-yb (changes requested):
The javadoc (and the PR description) say the tag is omitted under the embedded engine- that's not accurate. EmbeddedEngine defines name as a required field and passes the full config to connector.start(), our taskConfigs() copies all props into each task, so under the embedded engine / Debezium Server the tag is always emitted with the engine name. Our own test framework does this too (TestBaseClass sets name=test-connector), so all integration tests will now emit connector=test-connector.

The behavior is fine (arguably desirable), but please:

1. Fix the javadoc: the empty-map case only applies to hand-built Configurations (e.g. unit tests), not the embedded engine.
2. Update the PR description: embedded/Debezium Server metric names do change with this PR, worth calling out in release notes.

shishir2001-yb (commented):
LGTM

shishir2001-yb (approved):

PR #422, opened 2026-08-21
[DBZ][yugabyte/yugabyte-db#30831] Use max_index_in_sort_window while sending GetChnages rpcs

Sumukh-Phalgaonkar:
#### Problem

The diff https://phorge.dev.yugabyte.com/D56990  added a new field called `max_index_in_sort_window` to the GetChangesRequest and GetChangesResponse protos to solve a CDC data loss problem. However the gRPC connector cannot use these fields directly. Hence this PR.

#### Solution

This PR updates the yb-client version to `0.8.122-20260819.111006-1` so that we can use the new field in GetChanges rpcs. The usage of the field `max_index_in_sort_window` is identical to that of `wal_segment_index`. This means that for every tablet, the value of `max_index_in_sort_window` from the previous GetChanges response is passed on to the next request. The first request goes in with the default value 0.

#### Testing

No UT could be written since this field comes into action only when we have out of order applies in the WAL. There is no deterministic way to produce this in a UT.
The existing UTs in `YugabyteDBDatatypesTest` were run to ensure that there is no regression.

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
@@ -192,7 +192,8 @@ public class YugabyteDBConsistentStreamingSource extends YugabyteDBStreamingChan
                                             table, streamId, tabletId, cp.getTerm(), cp.getIndex(), cp.getKey(),
                                             cp.getWrite_id(), cp.getTime(), schemaNeeded.get(tabletId),
                                             taskContext.shouldEnableExplicitCheckpointing() ? tabletToExplicitCheckpoint.get(part.getId()) : null,
-                                            tabletSafeTime.getOrDefault(part.getId(), -1L), offsetContext.getWalSegmentIndex(part));
+                                            tabletSafeTime.getOrDefault(part.getId(), -1L), offsetContext.getWalSegmentIndex(part),
+                                            null /* getchangesRespMaxSizeBytes */, offsetContext.getMaxIndexInSortWindow(part));
                                 } catch (CDCErrorException cdcException) {
                                     // Check if exception indicates a tablet split.
                                     if (cdcException.getCDCError().getCode() == CdcService.CDCErrorPB.Code.TABLET_SPLIT) {
@@ -239,6 +240,7 @@ public class YugabyteDBConsistentStreamingSource extends YugabyteDBStreamingChan
                                             response.getResp().getSafeHybridTime());
                                     offsetContext.updateWalPosition(part, finalOpid);
                                     offsetContext.updateWalSegmentIndex(part, response.getWalSegmentIndex());
+                                    offsetContext.updateMaxIndexInSortWindow(part, response.getMaxIndexInSortWindow());
 
                                     tabletSafeTime.put(part.getId(), response.getResp().getSafeHybridTime());
 
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
@@ -50,6 +50,7 @@ public class YugabyteDBOffsetContext implements OffsetContext {
     private IncrementalSnapshotContext<TableId> incrementalSnapshotContext;
     private YugabyteDBConnectorConfig connectorConfig;
     private final Map<String, Integer> tabletWalSegmentIndex;
+    private final Map<String, Long> tabletMaxIndexInSortWindow;
 
     private YugabyteDBOffsetContext(YugabyteDBConnectorConfig connectorConfig,
                                     YugabyteDBTransactionContext transactionContext,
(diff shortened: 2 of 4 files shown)


src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
```diff
@@ -195,6 +199,14 @@ public Integer getWalSegmentIndex(YBPartition partition) {
         return this.tabletWalSegmentIndex.getOrDefault(partition.getId(), 0);
     }
 
+    public void updateMaxIndexInSortWindow(YBPartition partition, long maxIndexInSortWindow) {
+        this.tabletMaxIndexInSortWindow.put(partition.getId(), maxIndexInSortWindow);
```
shishir2001-yb, line 203: `handleTabletSplit` removes the split parent's entries from schemaNeeded and `tabletToExplicitCheckpoint`, but the new `tabletMaxIndexInSortWindow` entry is never removed, so the map grows unboundedly across splits for a long-lived task?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
```diff
@@ -239,6 +240,7 @@ protected void getChanges2(ChangeEventSourceContext context, YBPartition ybParti
                                             response.getResp().getSafeHybridTime());
                                     offsetContext.updateWalPosition(part, finalOpid);
                                     offsetContext.updateWalSegmentIndex(part, response.getWalSegmentIndex());
+                                    offsetContext.updateMaxIndexInSortWindow(part, response.getMaxIndexInSortWindow());
```
shishir2001-yb, line 243: It's inside `for (CDCSDKProtoRecordPB record : ...)`, do we need to also keep it outside as well? Are there any scenarios where we won't get any records? Not even a safepoint?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
```diff
@@ -195,6 +199,14 @@ public Integer getWalSegmentIndex(YBPartition partition) {
...
+    }
+
+    public Long getMaxIndexInSortWindow(YBPartition partition) {
+        return this.tabletMaxIndexInSortWindow.getOrDefault(partition.getId(), 0L);
```
shishir2001-yb, line 207: tabletMaxIndexInSortWindow is in-memory only i.e it's never written into getOffset(), so every task restart, rebalance, or worker failover sends max_index_in_sort_window=0 on the first GetChanges per tablet. Will this be an issue? 

Your YB PR says this
>  // We should never move the max_index_in_sort_window backwards as that could lead to us missing
  // out on sending data.

shishir2001-yb (changes requested):
max_index_in_sort_window or max_windo_in_sort_window?

Additionally, a lot of changes have come in ybclient since last release, maybe we want to run the whole suite?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
Sumukh-Phalgaonkar: `schemaNeeded` is a local map in getChnages function and `tabletToExplicitCheckpoint` is part of the `YugabyteDBStreamingChangeEventSource` class. We do not clear parent's entries for any of the existing maps that are owned by `YugabyteDBOffsetContext`.  Some of these maps like `tabletSourceInfo` are much more heavier, than `tabletMaxIndexInSortWindow`. In `tabletMaxIndexInSortWindow` we are only storing an Long per tablet. I dont know what is the reasoning behind not clearing `tabletSourceInfo` on tablet split, but I have kept the implementation consistent here.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBOffsetContext.java
Sumukh-Phalgaonkar: No this will not be an issue, since wal_segment_index is also not persisted. The YB diff comment is not relevant here, it is from the source / generation side.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
Sumukh-Phalgaonkar: No it is outside the loop. There is no `max_index_in_sort_window` per record, it is per response.

Sumukh-Phalgaonkar:
> max_index_in_sort_window or max_windo_in_sort_window?
> 
> Additionally, a lot of changes have come in ybclient since last release, maybe we want to run the whole suite?

It should be `max_index_in_sort_window`, changed the title and summary accordingly.

We can skip the whole suite, since the relevant change in ybclient is only the proto change I made. Other changes are from YBA side. Running one suite is enough to ensure no regressions. The LR tests anyway have already covered other scenarios like tablet split etc.

shishir2001-yb (approved):

PR #423, opened 2026-08-24
Make failure-path error messages actionable

yaronp68:
## Why

An audit of every `catch` / `throw` / error-log site under `src/main` turned up a consistent problem on the connector's failure paths: the message an operator gets is frequently too vague to act on, and in several places the underlying exception is discarded entirely, so the real cause survives only in a log line — or not at all.

Two of the messages were actively misleading. `"restart_lsn could be found"` was missing its negation and stated the opposite of what had happened. The master-connection failure reported the YSQL hostname rather than the master addresses it had actually tried to contact.

This PR covers the message-quality findings only. The separate class of findings — non-retriable errors that never reach the connector at all — is not addressed here; see the follow-ups below.

## What changed

- **Attach the cause** wherever a wrapping exception discarded it: the catch-up streaming stopping LSN, the replication-slot LSN reads, and the array value converter.
- **Add the identifiers needed to correlate against the server**: slot, plugin and database on the LSN failures; table and tablet on the schema refresh; stream and table IDs on the tablet fetch; the offending value on an unparseable or invalid LSN.
- **Fix the two misleading messages** described above.
- **Split the table poller's messages** by path. The DB-stream-info path and the publication path previously shared one message that named stream info even when the failure came from `pg_publication_tables`. The `"will retry again"` claim is now stated as what actually happens — a check on the next poll interval.
- **Stop the generic wrapper in `validateConnection`** from masking the specific validation messages underneath it. A `DebeziumException` already carrying a user-facing message is now rethrown as-is.
- **Log the throwable** rather than only its code or its `toString()` in the CDC error paths and in the error handler, so a stack trace is available.
- **Replace the four identical copies** of the COMMIT-without-BEGIN message with one helper that reports the partition, transaction, LSN and commit time, plus which of the two code paths raised it.
- **Throw instead of returning null** from `getCheckpointWithRetry`'s unreachable tail, so a bug there cannot surface as a message-less `NullPointerException` in the caller.

## Compatibility notes

The wraps in `fetchTableList` and `getTabletListMappedToTableIds` deliberately keep their cause-only form. `DebeziumException(Throwable)` copies the cause's `toString()` into the message, and `YugabyteDBConfigTest:169` and `:223` depend on that. Their added context goes to the log instead, leaving the exception chain unchanged.

The LSN parsing helpers in `YugabyteDBConnection` become package-private so their failure paths can be unit tested without a running server.

## Tests

These paths had no coverage. Both new test classes run without a YugabyteDB container, so they add seconds rather than minutes to the build.

- `YugabyteDBConnectionLsnMessageTest` drives the replication-slot LSN parsing through a `Proxy`-based `ResultSet` stub and asserts that each failure names the column, slot, plugin and database, quotes the offending value where there is one, and keeps the `SQLException` reachable as the cause. The `confirmed_flush_lsn` case also checks that the first failure survives as a suppressed exception once the `restart_lsn` fallback fails too. Two happy-path cases pin the behaviour for a null column and a well-formed LSN.
- `CommitWithoutBeginMessageTest` covers the extracted helper: that it identifies the partition, transaction, checkpoint and commit time, that it distinguishes the two call sites, and that a null transaction ID does not make message construction itself throw while it is reporting a failure.

10 tests, all passing locally.

## Follow-ups worth filing

1. **Negative-path assertions in `YugabyteDBConfigTest` cannot fail.** They sit inside `startEngine` completion callbacks, which run on an executor thread whose `Future` `TestBaseClass` discards, so an `AssertionError` raised there never reaches the JUnit thread. Repairing the harness is the highest-value test change available in this repo, but it will turn several currently-green tests red before they are fixed — which is why it is not in this PR.
2. **Four of those assertions test a path the embedded engine never executes.** `EmbeddedEngine` calls `SourceConnector.start(Map)` but never `SourceConnector.validate(Map)`, so `validateConnection` is not invoked in that test path, yet four tests assert on strings that exist only inside it. They need to call `connector.validate(...)` directly.
3. **Non-retriable errors that are never propagated** — a separate set of findings from the same audit, including a dead retry loop in `YugabyteDBTablePoller` that makes `fail()` unreachable, and a swallow-then-`requireNonNull` pattern in `YBClientUtils` that turns real errors into message-less NPEs.

### DIFF
--- a/src/test/java/io/debezium/connector/yugabytedb/connection/YugabyteDBConnectionLsnMessageTest.java
+++ b/src/test/java/io/debezium/connector/yugabytedb/connection/YugabyteDBConnectionLsnMessageTest.java
@@ -1,0 +1,184 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.yugabytedb.connection;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Proxy;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.Map;
+
+import org.apache.kafka.connect.errors.ConnectException;
+import org.junit.jupiter.api.Test;
+
+import io.debezium.config.Configuration;
+import io.debezium.jdbc.JdbcConfiguration;
+
+/**
+ * Unit tests for the failure paths of the replication-slot LSN parsing in
+ * {@link YugabyteDBConnection}. These paths only ever surface as an exception message in a
+ * connector log, so the message is the contract: it has to say which column could not be read,
+ * for which slot, and it has to keep the underlying {@link SQLException} as the cause.
+ * <p>
+ * No server is needed here - {@link YugabyteDBConnection} does not open a connection in its
+ * constructor, and the parsing methods take a {@link ResultSet} that is stubbed below.
+ *
+ * @author Yaron Parasol
+ */
+public class YugabyteDBConnectionLsnMessageTest {
+
+    private static final String SLOT = "test_slot";
+    private static final String PLUGIN = "yboutput";
+    private static final String DATABASE = "test_db";
+
+    /**
+     * A {@link ResultSet} stub that serves {@code getString(String)} from the supplied map. A null
+     * map value means the column is present but null; a column missing from the map throws a
+     * {@link SQLException}, which is how pgjdbc reports a column that is not in the result set.
+     * Every other method is unsupported - the code under test must not need them.
+     */
+    private static ResultSet resultSetReturning(Map<String, String> columns) {
+        InvocationHandler handler = (proxy, method, args) -> {
+            if ("getString".equals(method.getName()) && args != null && args.length == 1) {
+                String column = String.valueOf(args[0]);
+                if (!columns.containsKey(column)) {
+                    throw new SQLException("The column name " + column + " was not found in this ResultSet.");
+                }
+                return columns.get(column);
+            }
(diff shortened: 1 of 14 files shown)


src/test/java/io/debezium/connector/yugabytedb/connection/YugabyteDBConnectionLsnMessageTest.java
```diff
@@ -0,0 +1,184 @@
...
+        assertTrue(ex.getMessage().contains(PLUGIN), "Message should name the plugin, was: " + ex.getMessage());
+        assertTrue(ex.getMessage().contains(DATABASE), "Message should name the database, was: " + ex.getMessage());
+
+        // The JDBC failure must remain reachable, not just be logged.
+        assertTrue(ex.getCause() instanceof SQLException,
+                "The SQLException should be attached as the cause");
+    }
+
+    @Test
+    public void confirmedFlushLsnFailureShouldReportBothColumnsAndKeepBothCauses() {
```
pranjal82, line 107: Missing Test: Fallback to restart_lsn Works When confirmed_flush_lsn fails.

Test for: First column missing, second column present → fallback works and returns the value

pranjal82 (changes requested):

src/test/java/io/debezium/connector/yugabytedb/connection/YugabyteDBConnectionLsnMessageTest.java
yaronp68: Good catch — that was the case the method actually exists for, and I had only covered both columns failing.

Added in 703b08a:

- `shouldFallBackToRestartLsnWhenConfirmedFlushLsnCannotBeRead` — exactly what you asked for: `confirmed_flush_lsn` missing, `restart_lsn` present, asserts the returned Lsn is the `restart_lsn` value and is valid.
- `confirmedFlushLsnShouldBeUsedWhenItIsPresent` — the other side of the branch. Both columns readable and set to different values, asserting the primary one wins so the fallback cannot silently take over.
- `shouldFallBackToRestartLsnEvenWhenItIsNull` — `confirmed_flush_lsn` missing and `restart_lsn` present but null. `tryParseLsn` returns null for a null column rather than throwing, so the fallback succeeds with a null result, and that must not be reported as a failure.

I checked the new tests can actually fail: pointing the fallback at a wrong column name fails the two fallback tests and nothing else in the class. Whole class is 13 tests, green.

---
_automated · Claude Code (Opus 5)_

PR #424, opened 2026-08-31
Stop eagerly evaluating debug/trace log arguments

bdbene:
## Problem

SLF4J's `{}` placeholders defer `toString()` on an argument, but they do not defer *evaluation of the argument expression*. String concatenation with `+` isn't deferred at all. So a `LOGGER.debug(...)` whose argument does real work pays that cost on every call, at every log level, including when DEBUG is disabled.

The hot one is in `YugabyteDBConnectorTask.doPoll()`, which runs on every poll of every task:

```java
final List<DataChangeEvent> records = queue.poll();
LOGGER.debug("Got the records from queue: " + records);
```

That `+` stringifies the entire batch — every `DataChangeEvent`, every `SourceRecord`, every Kafka Connect `Struct` and its schema — and then throws the result away. `PGCompatible` and `StriimCompatible` do the same thing per record via `SchemaUtil.asString` / `asDetailedString`.

## Changes

Across 13 files:

- `LOGGER.debug("..." + x)` → `LOGGER.debug("... {}", x)` where the argument is a cheap reference or accessor.
- Wrapped in `isDebugEnabled()` / `isTraceEnabled()` where the argument expression itself does work — `SchemaUtil.asString`, `asDetailedString`, collection copies, stream collection.
- Two malformed calls fixed, where a `{}` placeholder was combined with `+` concatenation so the concatenated tail was built eagerly and the placeholder consumed only the first argument:

```java
-    LOGGER.debug("Received DDL message {}", message.getSchema().toString()
-            + " the table is " + message.getTable());
+    LOGGER.debug("Received DDL message {} the table is {}",
+            message.getSchema(), message.getTable());
```

Log message content is unchanged at DEBUG and TRACE.

## Measured impact

Measured on a staging deployment (~200 connector tasks, 13-17k records/s aggregate), comparing several hours of steady-state before the change against several hours after.

**Allocation** — the record-stringification path is gone. `YugabyteDBConnectorTask.doPoll` and `ConnectRecord.toString` were the #2 and #3 callers of `String.valueOf` (32.2% and 13.3% of its subtree); both have disappeared from the caller list entirely. What remains under `String.valueOf` is JMX metric naming. Its share of sampled TLAB allocation went from 95.75% to 0.40%.

Allocation per connector task dropped 41% (4.65 → 2.73 MB/s).

**GC** — Shenandoah went from 7 of the top 18 CPU frames to none, with −66% to −84% relative on every individual GC frame.

**String machinery** — `Struct.toString` −100%, `jbyte_disjoint_arraycopy_avx3` −84%, `AbstractStringBuilder.append` −82%, `ensureCapacityInternal` −63%, `String.getBytes` −60% (all as share of CPU).

**CPU** — total container CPU moved 6.44 → 6.11 cores, and per-task CPU dropped 17%. Both figures are confounded: the after window ran twice the pods (5 → 10), 14% more tasks (182 → 207) and 20% less throughput (16.9k → 13.4k records/s), so the aggregate CPU delta is not cleanly attributable to this change. The allocation and GC numbers above are the reliable ones. After the change the remaining profile is dominated by poll-loop timekeeping (`os::javaTimeMillis` → `clock_gettime`, ~28%), not string building.

### Flame graphs

Allocation profile before the change — the `doPoll` → `String.valueOf` → `DataChangeEvent.toString` → `Struct.toString` chain:

image

Same environment after the change:

image

## Deliberately not changed

Calls whose arguments are cheap accessors (`getTransactionId()`, `part.getId()`, `size()`, `getOid()`) are left as plain `{}` parameterization without a level guard — the guard would cost more than the argument.

## Testing

`mvn compile` passes. No behavioral change at INFO and above; at DEBUG and TRACE the emitted messages are identical to before.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
+++ b/src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
@@ -44,7 +44,9 @@ public class YugabyteDBConsistentStreamingSource extends YugabyteDBStreamingChan
     protected void getChanges2(ChangeEventSourceContext context, YBPartition ybPartition,
                                YugabyteDBOffsetContext offsetContext,
                                boolean previousOffsetPresent) throws Exception {
-        LOGGER.debug("The offset is " + offsetContext.getOffset());
+        if (LOGGER.isDebugEnabled()) {
+            LOGGER.debug("The offset is {}", offsetContext.getOffset());
+        }
         LOGGER.info("Processing consistent messages");
 
         try (YBClient syncClient = YBClientUtils.getYbClient(this.connectorConfig)) {
@@ -57,7 +59,7 @@ public class YugabyteDBConsistentStreamingSource extends YugabyteDBStreamingChan
             Set<String> tIds =
                     partitionRanges.stream().map(HashPartition::getTableId).collect(Collectors.toSet());
             for (String tId : tIds) {
-                LOGGER.debug("Table UUID: " + tIds);
+                LOGGER.debug("Table UUID: {}", tIds);
                 YBTable table = syncClient.openTableByUUID(tId);
                 tableIdToTable.put(tId, table);
 
@@ -67,7 +69,7 @@ public class YugabyteDBConsistentStreamingSource extends YugabyteDBStreamingChan
                 tabletListResponse.put(tId, resp);
             }
 
-            LOGGER.debug("The init tabletSourceInfo before updating is " + offsetContext.getTabletSourceInfo());
+            LOGGER.debug("The init tabletSourceInfo before updating is {}", offsetContext.getTabletSourceInfo());
 
             // Initialize the offsetContext and other supporting flags
             Map<String, Boolean> schemaNeeded = new HashMap<>();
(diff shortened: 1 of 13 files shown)


bdbene:
I've opted for putting guards around these logs, also open to just removing them or rewriting them such that they don't log the record's contents (not great practice from data confidentiality perspective).

shishir2001-yb:
Added this to my queue

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
```diff
@@ -57,7 +59,7 @@ protected void getChanges2(ChangeEventSourceContext context, YBPartition ybParti
             Set<String> tIds =
                     partitionRanges.stream().map(HashPartition::getTableId).collect(Collectors.toSet());
             for (String tId : tIds) {
-                LOGGER.debug("Table UUID: " + tIds);
+                LOGGER.debug("Table UUID: {}", tIds);
```
shishir2001-yb: This should be `tId`. It was wrong before as well

shishir2001-yb (changes requested):
Just one minor comment but looks good overall

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
bdbene: Good catch — fixed in f03c8a7. It was logging the whole set on every iteration; now logs `tId`. Only occurrence of that pattern in the repo.

shishir2001-yb:
@bdbenem,  Looks good. I will merge it next week after releasing a stable connector(Which was already validated before this PR).

shishir2001-yb (approved):

PR #428, opened 2026-09-11
Fail fast on non-retriable YugabyteDB CDC errors

pranjal82:
## Summary
- Classify CDC / master errors so clearly permanent failures fail the task instead of looping on max.connector.retries.
- Permanent fail-fast examples: CHECKPOINT_TOO_OLD, SUBSCRIBER_NOT_FOUND, OPERATION_DISALLOWED, AUTO_FLAGS_CONFIG_VERSION_MISMATCH, non-publication TABLE_NOT_FOUND, and INVALID_REQUEST / UNKNOWN / INTERNAL when AppStatus or message shows a permanent condition (deleted stream, invalid stream/tablet id, not in stream, etc.).
- Tablet-split handling only when the server reports a split: CDC TABLET_SPLIT, or AppStatus TABLET_SPLIT.   
- Bare INVALID_REQUEST is not treated as a split (aligned with main); it retries unless it is clearly permanent.
- TABLE_NOT_FOUND retries on a real publication path via task flag task.use.publication (survives stream-id injection). Plain gRPC stream tasks still fail fast.
- Unknown / future CDC codes default to RETRY for forward compatibility.
- Bootstrap-curable cases (e.g. stream expired for tablet before SetCheckpoint) retry so the enlarged bootstrap budget can be spent.

## Test plan
- [x] `mvn -Dtest=YugabyteDBCdcErrorClassifierTest test`
- [x] Local yugabyted: delete stream, drop table, remove table from stream fail fast; TEST GetChanges injects 0/1/3 retry
- [ ] CI unit tests
- [ ] Optional local service IT: `mvn -Dyb.service.it=true -Dyb.bin.dir=<yb-bin> -Dtest=YugabyteDBCdcServiceErrorIT test`
- [ ] Confirm dynamic table add on a gRPC stream does not fail the task on a brief `TABLE_NOT_FOUND`


pranjal82:
@shishir2001-yb could you review this PR? I do not have permission to add reviewers from a fork.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
+/*
+ * Copyright Debezium Authors.
```
shishir2001-yb, line 2: YugabyetDB Authors*

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+import org.yb.cdc.CdcService.CDCErrorPB;
+import org.yb.client.CDCErrorException;
+import org.yb.client.MasterErrorException;
+import org.yb.master.MasterTypes.MasterErrorPB;
+
+import io.debezium.DebeziumException;
+import io.debezium.config.Configuration;
+
+/**
+ * Classifies CDC and master errors before the generic connector retry loop.
```
shishir2001-yb, line 22: we don't need "master"

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+ */
+final class YugabyteDBCdcErrorClassifier {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(YugabyteDBCdcErrorClassifier.class);
+
+    enum CdcErrorAction {
+        HANDLE_IN_STREAM,
+        FAIL_FAST,
+        RETRY
```
shishir2001-yb, line 31: Add a short in line comment explaining each

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+                // Any publication deployment can see transient TABLE_NOT_FOUND while
+                // CDC metadata catches up (including autocreate.mode=disabled, where
+                // operators add tables by hand and the table poller reconfigures).
+                // Gate only on usePublication — not on autocreate mode.
+                return usePublication ? CdcErrorAction.RETRY : CdcErrorAction.FAIL_FAST;
+            case INVALID_REQUEST:
+                // main intentionally stopped treating bare INVALID_REQUEST as a
+                // tablet split (transient GetChanges failures use this code).
+                // Only HANDLE_IN_STREAM when AppStatus explicitly says TABLET_SPLIT;
+                // otherwise retry (or fail-fast for clearly permanent AppStatus).
```
shishir2001-yb, line 132: We no longer want to treat a INVALID_REQUEST as a TABLET_SPLIT. Refer to https://github.com/yugabyte/debezium-connector-yugabytedb/pull/429 for more details

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+            return false;
+        }
+        ErrorCode appStatusCode = error.getStatus().getCode();
+        return appStatusCode == ErrorCode.INVALID_ARGUMENT
+                || appStatusCode == ErrorCode.NOT_AUTHORIZED
+                || appStatusCode == ErrorCode.NOT_SUPPORTED
+                || appStatusCode == ErrorCode.CONFIGURATION_ERROR
+                || appStatusCode == ErrorCode.DELETED
+                || appStatusCode == ErrorCode.EXPIRED
+                || appStatusCode == ErrorCode.NOT_FOUND
```
shishir2001-yb, line 186: For status `ErrorCode.NOT_FOUND` are you sure there are no cases where we should be retrying?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+                || appStatusCode == ErrorCode.RUNTIME_ERROR;
```
shishir2001-yb, line 187: Same question as above

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    private static CdcErrorAction actionForUnknownOrInternal(CDCErrorPB error) {
+        // Only trust the AppStatus tablet-split code here — not message substrings.
+        if (error.hasStatus() && error.getStatus().getCode() == ErrorCode.TABLET_SPLIT) {
+            return CdcErrorAction.HANDLE_IN_STREAM;
```
shishir2001-yb, line 193: Is there any other places which could return us `ErrorCode.TABLET_SPLIT` and we were dependent on connector retries and handleTableSplit()? If yes, then with your change it will no longer work? Check it once

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+            default:
+                // Forward-compatible: a newer server may send codes this connector
+                // build does not know. Retry like the old generic catch loop.
+                return CdcErrorAction.RETRY;
+        }
+    }
+
+    private static CdcErrorAction actionForInvalidRequest(CDCErrorPB error) {
+        // Confirmed split via AppStatus — handle in stream.
+        if (error.hasStatus() && error.getStatus().getCode() == ErrorCode.TABLET_SPLIT) {
```
shishir2001-yb, line 157: Just pasting from below

>Is there any other places which could return us `ErrorCode.TABLET_SPLIT` and we were dependent on connector retries and handleTableSplit()? If yes, then with your change it will no longer work? Check it once

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+                return true;
+            default:
+                return false;
+        }
+    }
+
+    private static boolean hasFatalMessage(CDCErrorPB error) {
+        // Do not treat "is expired for tablet" as fatal: that message is also returned when
+        // SetCheckpoint has not been issued yet and is cured by makeStreamActive / bootstrap
+        // retries (see YugabyteDBSnapshotChangeEventSource#isSnapshotRequired).
```
shishir2001-yb, line 227: Do we need this comment? I don't think so

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+        return containsAny(statusMessage(error),
+                "could not find cdc stream",
+                "is not part of stream",
+                "not found under stream");
```
shishir2001-yb, line 231: Cross check if there is any case where server may return this but it should be retriable

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+                return false;
+        }
+    }
+
+    private static boolean hasFatalMessage(CDCErrorPB error) {
+        // Do not treat "is expired for tablet" as fatal: that message is also returned when
+        // SetCheckpoint has not been issued yet and is cured by makeStreamActive / bootstrap
+        // retries (see YugabyteDBSnapshotChangeEventSource#isSnapshotRequired).
+        return containsAny(statusMessage(error),
+                "could not find cdc stream",
```
shishir2001-yb, line 229: Same as below

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+                return true;
+            case INVALID_ARGUMENT:
+            case RUNTIME_ERROR:
+            case NOT_AUTHORIZED:
+            case NOT_SUPPORTED:
+            case CONFIGURATION_ERROR:
+            case DELETED:
+            case EXPIRED:
+            case NOT_FOUND:
+                return false;
```
shishir2001-yb, line 257: Have asked the same question couple of comments back

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    static boolean isFatalMasterError(Throwable error) {
+        Throwable current = error;
+        while (current != null) {
+            if (current instanceof MasterErrorException) {
+                MasterErrorException masterError = (MasterErrorException) current;
+                // Only trust explicit permanent master codes. Free-text matching
+                // ("does not exist") is too broad: stale tserver-cache lookups during
+                // bootstrap can surface similar wording and must still be retried.
```
shishir2001-yb, line 270: Not needed

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+                return masterError.error != null && isFatalMasterCode(masterError.error.getCode());
+            }
+            current = current.getCause();
+        }
+        return false;
+    }
+
+    private static boolean isFatalMasterCode(MasterErrorPB.Code code) {
+        // INVALID_REQUEST is deliberately omitted: it can appear from transient
+        // master/tserver cache inconsistency and should use the connector retry budget.
```
shishir2001-yb, line 280: Remove

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConnectorConfig.java
```diff
@@ -778,6 +778,18 @@ public JdbcConfiguration getJdbcConfig() {
...
+     * Set by {@link YugabyteDBgRPCConnector} on both the connector-scoped config (after
+     * stream-id injection in {@code start()}) and on task props. Needed because
+     * {@link #shouldUsePublication(Configuration)} is only valid before the resolved
+     * stream id is written into config.
+     */
+    public static final Field TASK_USE_PUBLICATION = Field.create(TASK_CONFIG_PREFIX + "use.publication")
+            .withDisplayName("Task uses publication")
+            .withType(Type.BOOLEAN)
+            .withDefault(false)
+            .withDescription("Internal config: true when the connector was started with the publication/slot path.");
```
shishir2001-yb, line 791: Wondering if it is even needed

shishir2001-yb: With the current thing user can explicitly set this in connector config. Is there precedence?

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBConsistentStreamingSource.java
```diff
@@ -197,12 +197,11 @@ protected void getChanges2(ChangeEventSourceContext context, YBPartition ybParti
                                             tabletSafeTime.getOrDefault(part.getId(), -1L), offsetContext.getWalSegmentIndex(part),
                                             null /* getchangesRespMaxSizeBytes */, offsetContext.getMaxIndexInSortWindow(part));
                                 } catch (CDCErrorException cdcException) {
-                                    // Check if exception indicates a tablet split.
-                                    if (cdcException.getCDCError().getCode() == CdcService.CDCErrorPB.Code.TABLET_SPLIT) {
+                                    // Consistent/transaction-ordering path historically handled
+                                    // TABLET_SPLIT only. Do not use HANDLE_IN_STREAM here — that
+                                    // could widen to other codes; gate on split evidence alone.
+                                    if (YugabyteDBCdcErrorClassifier.isTabletSplit(cdcException.getCDCError())) {
```
shishir2001-yb, line 203: This is already done in latest code. Rebase your PR and remove this. Refer to https://github.com/yugabyte/debezium-connector-yugabytedb/pull/429

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBgRPCConnector.java
```diff
@@ -224,6 +224,7 @@ public List<Map<String, String>> taskConfigs(int maxTasks) {
             }
 
             taskProps.put(YugabyteDBConnectorConfig.STREAM_ID.toString(), streamIdValue);
+            taskProps.put(YugabyteDBConnectorConfig.TASK_USE_PUBLICATION.toString(), String.valueOf(usePublication));
```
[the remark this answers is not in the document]
shishir2001-yb: How did you resolve this comment? It's a good practise to reply to each comment and explain how you have fixed it

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBSnapshotChangeEventSource.java
```diff
@@ -878,6 +880,7 @@ public void markSnapshotDoneOnServer(YBPartition partition, YugabyteDBOffsetCont
         // Break upon successful request.
         break;
       } catch (Exception e) {
+        YugabyteDBCdcErrorClassifier.throwIfFailFast(e, connectorConfig);
```
shishir2001-yb, line 883: Cross check there is no other place which you have missed. Might not be `catch (Exception e)` but could be a specific error

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -572,13 +578,14 @@ protected void getChanges2(ChangeEventSourceContext context,
                                     TEST_explicitCheckpoints.put(tabletId, explicitCheckpoint);
                                 }
                             } catch (CDCErrorException cdcException) {
-                                // Check if exception indicates a tablet split.
+                                // Tablet splits use CDCErrorPB.TABLET_SPLIT (or AppStatus TABLET_SPLIT).
+                                // Do not treat bare INVALID_REQUEST as a split — that can be a
+                                // transient GetChanges failure and must retry (see main).
```
shishir2001-yb, line 583: Undo your change

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+                    cdcException.getCDCError().hasStatus()
+                            ? cdcException.getCDCError().getStatus()
+                            : "none",
+                    error);
+        }
+        else {
+            LOGGER.error("Failing fast for non-retriable YugabyteDB error", error);
+        }
+
+        throw error instanceof Exception ? (Exception) error : new DebeziumException(error);
```
shishir2001-yb, line 91: Every caller is inside `catch (Exception e)`, so error is always an Exception? The else new DebeziumException(error) branch is dead. Maybe we can drop the DebeziumException import. If not, explain why

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+                        config.getString(YugabyteDBConnectorConfig.PUBLICATION_AUTOCREATE_MODE),
+                        YugabyteDBConnectorConfig.DEFAULT_PUBLICATION_AUTOCREATE_MODE);
+        return actionFor(error, publicationMode, YugabyteDBConnectorConfig.usesPublication(config));
+    }
+
+    static CDCErrorException findCdcError(Throwable error) {
+        Throwable current = error;
+        while (current != null) {
+            if (current instanceof CDCErrorException) {
+                return (CDCErrorException) current;
```
shishir2001-yb, line 58: This is pre-java 16. Can be replace with below. Note: We use Java 17
```
if (current instanceof CDCErrorException cdcError) {
    return cdcError;
}
```

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+            }
+            current = current.getCause();
+        }
```
shishir2001-yb, line 61: This and another loop in some function below is unnecessary. 
Can be replaced with below. Cross check 
```
import org.apache.commons.lang3.exception.ExceptionUtils;

static CDCErrorException findCdcError(Throwable error) {
    return ExceptionUtils.throwableOfType(error, CDCErrorException.class);
}

static boolean isFatalMasterError(Throwable error) {
    MasterErrorException masterError = ExceptionUtils.throwableOfType(error, MasterErrorException.class);
    return masterError != null
            && masterError.error != null
            && isFatalMasterCode(masterError.error.getCode());
}
```

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+     */
+    static CdcErrorAction actionFor(CDCErrorPB error,
+                                    YugabyteDBConnectorConfig.AutoCreateMode publicationMode) {
+        return actionFor(error, publicationMode,
+                publicationMode != YugabyteDBConnectorConfig.AutoCreateMode.DISABLED);
+    }
+
+    static CdcErrorAction actionFor(CDCErrorPB error,
+                                    YugabyteDBConnectorConfig.AutoCreateMode publicationMode,
+                                    boolean usePublication) {
```
shishir2001-yb, line 118: Both actionFor overloads parse `PUBLICATION_AUTOCREATE_MODE` and pass publicationMode into a method that never reads it. Remove the parse and the parameter.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    /**
+     * Used by unit tests. {@code FILTERED} and {@code ALL_TABLES} are treated as publication paths.
+     */
+    static CdcErrorAction actionFor(CDCErrorPB error,
+                                    YugabyteDBConnectorConfig.AutoCreateMode publicationMode) {
+        return actionFor(error, publicationMode,
+                publicationMode != YugabyteDBConnectorConfig.AutoCreateMode.DISABLED);
+    }
```
shishir2001-yb, line 114: Test-only overload in production code, and it encodes a different rule (mode != DISABLED) than production (usePublication flag). Both ITs call it with DISABLED hardcoded, so they never exercise the real entry point. Move it into the test.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
shishir2001-yb, line 132: Comment "transient GetChanges failures use this code" is false, every GetChanges INVALID_REQUEST is InvalidArgument (cdc_service.cc:1651, 1657, 1727, 1674).

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+            case NOT_LEADER:
+            case NOT_RUNNING:
+            case LEADER_NOT_READY:
+                return CdcErrorAction.RETRY;
+            case UNKNOWN_ERROR:
+            case INTERNAL_ERROR:
+                return actionForUnknownOrInternal(error);
+            default:
+                // Forward-compatible: a newer server may send codes this connector
+                // build does not know. Retry like the old generic catch loop.
```
shishir2001-yb, line 150: `default: RETRY` is unreachable for new server codes. `cdc_service.proto` is proto2 an unknown enum value is dropped to unknown fields and getCode() returns the field default UNKNOWN_ERROR. New codes land in `actionForUnknownOrInternal`. Fix the comment and PR description, add a WARN when !error.hasCode().

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        // Transient transport/leader issues — retry.
+        if (isExplicitlyTransientAppStatus(error)) {
+            return CdcErrorAction.RETRY;
+        }
+
+        // Permanent request failures — fail fast.
+        if (error.hasStatus() && (isFatalAppStatus(error) || hasFatalMessage(error))) {
+            return CdcErrorAction.FAIL_FAST;
```
shishir2001-yb, line 168: error.hasStatus() && is redundant both callees handle a missing status.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+                || appStatusCode == ErrorCode.NOT_AUTHORIZED
+                || appStatusCode == ErrorCode.NOT_SUPPORTED
+                || appStatusCode == ErrorCode.CONFIGURATION_ERROR
+                || appStatusCode == ErrorCode.DELETED
+                || appStatusCode == ErrorCode.EXPIRED
+                || appStatusCode == ErrorCode.NOT_FOUND
+                || appStatusCode == ErrorCode.RUNTIME_ERROR;
+    }
+
+    private static CdcErrorAction actionForUnknownOrInternal(CDCErrorPB error) {
```
shishir2001-yb, line 190: 155-173 vs 190-203. Order differs: transient-before-message here, message-before-transient there. Pick one.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+        }
+        switch (error.getStatus().getCode()) {
+            case SERVICE_UNAVAILABLE:
+            case TIMED_OUT:
+            case ABORTED:
+            case TRY_AGAIN_CODE:
+            case BUSY:
+            case LEADER_NOT_READY_TO_SERVE:
+            case LEADER_HAS_NO_LEASE:
+            case CACHE_MISS_ERROR:
```
shishir2001-yb, line 217: 175-188, 205-222, 234-261. 
Three functions, one decision. The fatal list at 180-187 and the false list at 249-256 are the same eight codes, so `isTransientAppStatus(e) == !isFatalAppStatus(e)` for every input, and the explicit transient cases only restate the default. One `EnumSet<ErrorCode> FATAL_APP_STATUS`.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+                || appStatusCode == ErrorCode.NOT_SUPPORTED
+                || appStatusCode == ErrorCode.CONFIGURATION_ERROR
+                || appStatusCode == ErrorCode.DELETED
+                || appStatusCode == ErrorCode.EXPIRED
+                || appStatusCode == ErrorCode.NOT_FOUND
+                || appStatusCode == ErrorCode.RUNTIME_ERROR;
+    }
+
+    private static CdcErrorAction actionForUnknownOrInternal(CDCErrorPB error) {
+        // Only trust the AppStatus tablet-split code here — not message substrings.
```
shishir2001-yb, line 191: Remove

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
```diff
@@ -0,0 +1,300 @@
...
+        return error.hasStatus() ? error.getStatus().getMessage().toLowerCase(Locale.ROOT) : "";
+    }
+
+    private static boolean containsAny(String value, String... tokens) {
+        for (String token : tokens) {
+            if (value.contains(token)) {
+                return true;
+            }
+        }
+        return false;
```
shishir2001-yb, line 298: StringUtils.containsAny(value, tokens) from commons-lang3.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
```diff
@@ -179,6 +178,10 @@ public void execute(ChangeEventSourceContext context, YBPartition partition, Yug
         }
     }
 
+    protected void failFastIfNonRetriableCdcError(Throwable error) throws Exception {
```
shishir2001-yb, line 181: Two idioms for the same thing: this wrapper here and in the consistent source, direct throwIfFailFast calls in the snapshot source and YBClientUtils.

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
```diff
@@ -433,6 +433,7 @@ public static GetTabletListToPollForCDCResponse getTabletListToPollForCDCWithRet
 
         return resp;
       } catch (Exception e) {
+        YugabyteDBCdcErrorClassifier.throwIfFailFast(e, connectorConfig);
```
shishir2001-yb, line 436: `getTabletListToPollForCDCWithRetry` also runs on the connector thread in taskConfigs (YugabyteDBgRPCConnector.java:400?), where a fail-fast now fails the connector immediately. Maybe Intended?

shishir2001-yb (changes requested):
I am yet to take a look at the test but until then you can work on the reaming comments. Also don't resolve the comment without explaining how you resolved it

src/main/java/io/debezium/connector/yugabytedb/YBClientUtils.java
pranjal82: Yes, intended. getTabletListToPollForCDCWithRetry discovers tablets for a stream/table and is used in both connector-side task assignment  and task-side streaming/snapshot paths

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: AppStatus NOT_FOUND is used for transient CDC failures as well as missing-stream cases: not-leader, tablet/peer not on this tserver, tablet leader lookup empty,replica-identity / metadata lag which should be retriable. removing NOT_FOUND from fatal APP STATUS codes.
NOT_FOUND retries, except when the CDC code is already fail-fast (TABLE_NOT_FOUND on gRPC, etc.) or the AppStatus message is a known missing-stream / not-in-stream string.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: Agreed, every caller is catch (Exception e), so the DebeziumException wrap is unreachable. 
I’ll take Exception, rethrow it, and drop the unused import.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: comment updated.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: as per my understanding, there is no other RPC that both returns ErrorCode.TABLET_SPLIT and depended on connector retries to reach handleTabletSplit(). The AppStatus TABLET_SPLIT check is a fallback; current GetChanges already sends CDC TABLET_SPLIT

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: Both paths now use the same order: split → fatal message → transient → otherwise

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: There is no other RPC that both sent TABLET_SPLIT and depended on connector retries to reach handleTabletSplit(). This change does not take that away

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: No current retriable miss. The server uses that text only on GetTabletListToPollForCDC, with CDC code TABLE_NOT_FOUND. We classify that code first: publication → retry, gRPC → fail-fast. The message list is not used.

It would only fail-fast wrongly if the CDC code were dropped and we saw UNKNOWN/INTERNAL plus that string. That pairing does not exist on the server today.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: No retriable GetChanges case. That text comes from master catalog (GetCDCStream / stream lookup): stream id missing, slot name unknown, or stream deleting.

That is a bad/deleted stream, not tablet/leader movement. Fail-fast seems correct.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: removed master related code

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: added inline comments for each

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: got it, updated code so that INVALID_REQUEST is never a tablet split: we fail-fast only on a fatal AppStatus or missing-stream message, otherwise we retry so a real split can come back as CDC TABLET_SPLIT.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: removed the comment

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBStreamingChangeEventSource.java
pranjal82: failFastIfNonRetriableCdcError was only a pass-through. Dropped it; every path now calls YugabyteDBCdcErrorClassifier.throwIfFailFast.

src/main/java/io/debezium/connector/yugabytedb/YugabyteDBCdcErrorClassifier.java
pranjal82: noted

